PR #180, opened 2021-04-16
Service Loader test case is not correct

### LINKED ISSUE #49: Service Loader test case is not correct
Original bug ID: [BZ#3128](https://osgi.org/members/bugzilla/show_bug.cgi?id=3128)
From: Georgi Boyvalenkov &lt;<georgi.boyvalenkov@bosch-si.com>&gt;
Reported version: R6

bosschaert:
Service Loader test testServiceClientWithFilter which uses
cardinality:=multiple should allow for 1 or 2 wires as
cardinality:=multiple is not required to provide all wires.

This fixes #49

### DIFF
--- a/org.osgi.test.cases.serviceloader/src/org/osgi/test/cases/serviceloader/junit/ServiceLoaderServiceClientTest.java
+++ b/org.osgi.test.cases.serviceloader/src/org/osgi/test/cases/serviceloader/junit/ServiceLoaderServiceClientTest.java
@@ -18,6 +18,8 @@
 
 package org.osgi.test.cases.serviceloader.junit;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


org.osgi.test.cases.serviceloader/src/org/osgi/test/cases/serviceloader/junit/ServiceLoaderServiceClientTest.java
```diff
@@ -213,7 +213,9 @@ public void testServiceClientWithFilter() throws Exception {
     				wires = rev.getWiring().getRequiredWires("osgi.serviceloader");
     				assertNotNull(wires);
 
-    				assertEquals("There are two providers so there should be two wires", 2, wires.size());
+					assertTrue(
```
bjhargrave, line 216: Can we do

```java
import static org.assertj.core.api.Assertions.assertThat;

assertThat(wires).as("There should be 2 wires as there are 2 providers, but 1 wire is also allowed since cardinality:=multiple is not guaranteed to provide all wires")
    .hasSizeBetween(1, 2);
```

This reads better and will give a more meaningful exception if the size is wrong.

bjhargrave (changes requested):

bjhargrave (approved):

Issue #155, opened 2021-03-08
[Initial Provisioning] Design flaw in use of Zip Extra field


bjhargrave:
In the [Initial Provisioning spec](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.provisioning.html), there is discussion of string the mime type of an entry in the Zip Extra field (see section 4.5 "Extensible data fields"of <https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT>).

This discussion is in the javadoc for [`ProvisioningService.addInformation(ZipInputStream)`](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.provisioning.html#org.osgi.service.provisioning.ProvisioningService.addInformation-ZipInputStream-) and the MIME constants in [ProvisioningService](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.provisioning.html#org.osgi.service.provisioning.ProvisioningService).

However, the specification fails to document the Zip Header ID value to contain the mime information. The Zip spec requires the extra field in a format with defined header id. The current RI (https://github.com/eclipse/rt.equinox.bundles/tree/master/bundles/org.eclipse.equinox.ip) and Bnd (prior to Bnd 5.3) just assumed the entire extra field held only the UTF-8 encoding of the mime type. But this is invalid and does not allow for multiple parties to store extra information.

The Initial Provisioning spec needs to be updated to declare a Header ID value to use for this information or to deprecate using the extra field and only support the replacement [InitialProvisioning-Entries](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.provisioning.html#i1442770) manifest header. The latter is probably the best choice as encoding a custom Zip extra field entry is hard to do without custom Zip tooling. For example, Bnd would need to be updated to support whatever Header Id the Initial Provisioning spec selects.

bjhargrave:
Another option is to remove the Initial Provisioning spec from Compendium for R8 if we deem it no longer of value to continue publishing.

bjhargrave:
See past discussion in https://github.com/osgi/bugzilla-archive/issues/1182

bjhargrave:
Project call: Agreed to remove IP Spec from R8.

bjhargrave:
> Project call: Agreed to remove IP Spec from R8.

[...]

PR #191, opened 2021-04-22
Remove Initial Provisioning and JMX from Compendium R8

bjhargrave:
Fixes https://github.com/osgi/osgi/issues/155


bjhargrave:
@osgi/osgi-committers Are we also ok with no longer adding JMX to Compendium R8? It was not in Compendium R7 anyway and its continued presence in Compendium R8 requires we also retain Initial Provisioning.

rotty3000 (approved):
LGTM

PR #193, opened 2021-04-27
Remove section about feature merging

bosschaert:


bjhargrave:
Is feature merging no longer a requirement of the specification?

Issue #187, opened 2021-04-20
Feature implementation development


bjhargrave:
Features design work is already complete prior to the new process. So we now open tracking issues for the specification, implementation, and tck work.

_Originally posted by @bjhargrave in https://github.com/osgi/osgi/issues/162#issuecomment-823601652_

bjhargrave:
Now that we have an RC1 build of org.osgi.service.feature in Maven Central, when do we expect an RC/Milestone build of the compatible implementation in Maven Central?

bosschaert:
org,apache.felix:org.apache.felix.feature:0.9.2-RC2 passes all the TCK tests. The Apache Felix vote has been kicked off on Tuesday 7 September. All going well this vote will conclude on Friday 10 September at which point it will be available from Maven Central.
https://www.mail-archive.com/dev%40felix.apache.org/msg52550.html

Until then https://repo1.maven.org/maven2/org/apache/felix/org.apache.felix.feature/0.9.0-RC1/ is already available in Maven Central. This implementation has a few test failures that are addressed in RC2.

PR #195, opened 2021-04-29
fix check in ID:fromMavenID

stbischof:
ref: https://github.com/osgi/osgi/issues/187


bosschaert (approved):
Well spotted! Thanks @stbischof !

PR #196, opened 2021-04-30
Additional feature attributes

### LINKED ISSUE #179: [Features] Provide same MetaData-Fields as Bundle
It would be nice if features would use same Metadata-Fields as Bundle and Subsystem.

|Bundle | Subsystem | Feature
| ------------- |:-------------:| -----:|
|Bundle-Category | Subsystem-Category | ?
|Bundle-ContactAddress | Subsystem-ContactAddress | ?
|Bundle-Copyright | Subsystem-Copyright | ?
|Bundle-Description | Subsystem-Description | description
|Bundle-Developers | ? | ?
|Bundle-DocURL | Subsystem-DocURL | ?
|Bundle-Icon | Subsystem-Icon |
|Bundle-License | Subsystem-License | license
|Bundle-Name | Subsystem-Name | title (maybe rename)
|Bundle-SCM | | ?
|Bundle-SymbolicName | Subsystem-SymbolicName | ?(ArtifactId)
|Bundle-Vendor | Subsystem-Vendor | vendor
|Bundle-Version | Subsystem-Version | version

bosschaert:
Added: categories, copyright, docURL and SCM.
This closes #179

### DIFF
--- a/org.osgi.util.feature/src/org/osgi/util/feature/Feature.java
+++ b/org.osgi.util.feature/src/org/osgi/util/feature/Feature.java
@@ -17,8 +17,6 @@
  *******************************************************************************/
 package org.osgi.util.feature;
 
-import java.net.URI;
-import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
@@ -41,7 +39,7 @@ public interface Feature extends FeatureArtifact {
 	 * 
 	 * @return The categories.
 	 */
-	Collection<String> getCategories();
+	List<String> getCategories();
 
 	/**
 	 * Get the copyright.
@@ -62,7 +60,7 @@ public interface Feature extends FeatureArtifact {
 	 * 
 	 * @return The documentation URL.
 	 */
-	URI getDocURL();
+	String getDocURL();
 
     /**
      * Get the vendor.
--- a/org.osgi.util.feature/src/org/osgi/util/feature/Feature.java
+++ b/org.osgi.util.feature/src/org/osgi/util/feature/Feature.java
@@ -17,6 +17,8 @@
  *******************************************************************************/
 package org.osgi.util.feature;
 
+import java.net.URI;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


org.osgi.util.feature/src/org/osgi/util/feature/Feature.java
```diff
@@ -35,35 +37,63 @@
...
 
     /**
-     * Get the description.
-     * @return The description.
-     */
+	 * Get the categories.
+	 * 
+	 * @return The categories.
+	 */
+	Collection<String> getCategories();
```
bjhargrave, line 44: Perhaps this should be `List` since there is an order in the document? I think `List` is the best mapping to a JSON array.

org.osgi.util.feature/src/org/osgi/util/feature/Feature.java
```diff
@@ -35,35 +37,63 @@
...
+	 * @return The description.
+	 */
     String getDescription();
 
+	/**
+	 * Get the documentation URL.
+	 * 
+	 * @return The documentation URL.
+	 */
+	URI getDocURL();
```
bjhargrave, line 65: This should probably just be `String` and let the user decide whether to parse into a URI or not.

org.osgi.util.feature/src/org/osgi/util/feature/FeatureBuilder.java
```diff
@@ -27,6 +28,45 @@
...
+	 */
+	FeatureBuilder setDescription(String description);
+
+	/**
+	 * Set the documentation URL.
+	 * 
+	 * @param docURL The Documentation URL.
+	 * @return This builder.
+	 */
+	FeatureBuilder setDocURL(URI docURL);
```
bjhargrave, line 69: As above, I think this should just be a `String`.

osgi.specs/docbook/708/util.features.xml
```diff
@@ -240,13 +241,27 @@ Features may have dependencies on other Features. Features inherit the capabilit
...
             <row>
               <entry>description</entry>
               <entry>String</entry>
               <entry>Optional</entry>
               <entry>A longer description of the Feature.</entry>
             </row>
+            
+            <row>
+              <entry>docURL</entry>
+              <entry>URI</entry>
```
bjhargrave, line 261: Is `URI` a JSON type? I would assume the type here is `String` and, as commented above, let the user decide whether to parse into a URI.

bjhargrave (changes requested):

stbischof:
Maybe the getter of the optional fields should return a Optional<String>.

org.osgi.util.feature/src/org/osgi/util/feature/Feature.java
bosschaert: Categories don't have an order but sets aren't supported in JSON, so I picked Collection to make the Java API closer to the semantics. However it's probably better to map the Java API closer to the JSON representation, so I've changed is as you suggested.

PR #197, opened 2021-04-30
Remove API relating to Feature merging.

bosschaert:


stbischof:
I am surprised, that feature merging no longer a requirement of the specification.

Or is it just because an external tooling should solve this outside of the scope of the spec?

bosschaert:
Yes, feature merging is really a tooling concern. I think we should keep the feature spec right now focused on the model itself. In the future we might have additional specs that build on top of the model, those specs could look at other things like merging features, launching features etc...

stbischof:
Ah okay. Expected that the Launcher would be part of the spec. But sure . Model is a good first step.

PR #212, opened 2021-05-14
promise: new PromiseFactory.resolvedWith(Promise) method

bjhargrave:
This allows the user to create a new Promise using a specific
PromiseFactory that is resolved with the specified promise. If the
user is using multiple promise factories to manage callback thread
usage, this method provides an efficient way to chain a promise
from one promise factory to a new promise from another promise
factory.


bjhargrave:
@pkriens I think this should address our discussion today about the new promise factory method.

PR #229, opened 2021-05-28
Change Features from util into a service spec

### LINKED ISSUE #216: [Features Impl] Handle external dependencies while being usable in both an OSGi and non-OSGi runtime
For the Features spec implementation we have the following constraints.

1. The implementation must be able to do the following

* Parse JSON
* Parse Configurator JSON (which is a specialized JSON format)

The best way to handle these points above is to be able to use existing implementations of these.

2. Furthermore, the Features spec implementation needs to be usable from both inside OSGi as well as in a plain Java environment. This is because commandline tools, or build/developer tooling plugins must be able to work with features, in addition to OSGi runtime based environments.

bosschaert:
Some text in the spec may need to be tweaked further for this, but at least this builds without errors and the spec appears at the right place in the Compendium book.

This fixes #216


osgi.specs/docbook/cmpn/001/introduction.xml
```diff
@@ -662,6 +662,17 @@
               linkend="service.log.stream"/></code></entry>
             </row>
 
+            <row>
+              <entry><xref linkend="service.features"
```
stbischof, line 666: may use singular `feature` not `features`

bosschaert: I think singular would then map to 'Feature Service Specification' which I think is a bit too general and potentially confusing. I would like to leave the name at 'Features Service Specification' which would then map to `service.features`.

stbischof:
lgtm.
But as the DCO app said, the commit is not sign off.

bosschaert:
> lgtm.
> But as the DCO app said, the commit is not sign off.

I thought that a 'sign off' in the commit was no longer necessary, but I could be wrong.

stbischof:
Maybe there is no need for members.
Then the DCO app has an option:
```
require:
  members: false
```

bjhargrave:
> I thought that a 'sign off' in the commit was no longer necessary, but I could be wrong.

But we do not have the ECA check installed on this repo yet, so I am requiring DCO for now. Just to cover the bases!

bjhargrave (changes requested):
Just need the signed off and its good to go.

bosschaert:
👍  I've added the signoff

bjhargrave (approved):

PR #230, opened 2021-06-01
[comp_tck] ConstructorInjection of unavalible optional Reference

stbischof:
Test the sentence about - [Injection of optional References in Constructor:](https://osgi.github.io/osgi/cmpn/service.component.html#service.component-constructor.injection)
```
If the constructor parameter is associated with a reference having cardinality of 0..1 
and there is no bound service for the reference, then the value null will be supplied as 
the constructor parameter. 
```

Felix fixed this issue in newest versions `2.1.26`

@bjhargrave  would it be okay, to remove the comments and update the version on this position?

https://github.com/osgi/osgi/blob/bf2fd917161748cdaad275d0e90c4d400055d811/cnf/ext/central.mvn#L104-L105


bjhargrave (approved):

tjwatson:
I tested this change along with PR #232 and the updated test passes with the latest Felix SCR for R8.

tjwatson (approved):

PR #234, opened 2021-06-03
converter: Use ParameterizedTest

bjhargrave:


org.osgi.test.cases.converter/src/org/osgi/test/cases/converter/felix/ConverterTest.java
```diff
@@ -429,38 +431,44 @@ public void testCharArrayConversion() {
...
-		checkArray(byte[][][].class);
-		checkArray(short[][][].class);
-		checkArray(char[][][].class);
-		checkArray(int[][][].class);
-		checkArray(float[][][].class);
-		checkArray(long[][][].class);
-		checkArray(double[][][].class);
+	@ParameterizedTest(name = "arrayType=\"{0}\"")
+	@ValueSource(classes = {
+			String[].class, //
```
bosschaert, line 436: I guess the empty comment at the end of each line is to prevent the formatter that always runs from collapsing all the values on a single line?

bosschaert (approved):

PR #235, opened 2021-06-03
Use of Optional<> in Features API.

### LINKED ISSUE #207: [Features] Use Optional-API in Features
There are several optional fields in Features.
This are the methods where an `Optional<String>` as return value could be a good way to go.

```
Features::getName
Features::getDescription
Features::getVendor
Features::getLicense
Features::getName
Features::getDocURL
Features::getCopyright
Features::getCategories
Features::getSCM
ID::getType
ID::getClassifier
FeatureConfiguration::getFactoryPid
```

Extensions: https://osgi.github.io/osgi/cmpn/util.features.html#d0e167706
At the moment this methods also would return null when used in combination with the wrong `type - getType`

This could also handled by an Optional. An other way for this would be to throw a `UnsupportedOperationException`
``` 
FeatureExtension::getJSON	(Optional<Stirng>)
FeatureExtension::getText	(Optional<List<Stirng>>)
FeatureExtension::getArtifacts	(Optional<List<ID>>)
```

bosschaert:
This fixes #207

### DIFF
--- a/org.osgi.service.feature/src/org/osgi/service/feature/Feature.java
+++ b/org.osgi.service.feature/src/org/osgi/service/feature/Feature.java
@@ -13,12 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- * SPDX-License-Identifier: Apache-2.0 
+ * SPDX-License-Identifier: Apache-2.0
  *******************************************************************************/
 package org.osgi.service.feature;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import org.osgi.annotation.versioning.ProviderType;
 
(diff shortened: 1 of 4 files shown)


org.osgi.service.feature/src/org/osgi/service/feature/Feature.java
```diff
@@ -13,14 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- * SPDX-License-Identifier: Apache-2.0 
+ * SPDX-License-Identifier: Apache-2.0
  *******************************************************************************/
 package org.osgi.service.feature;
 
+import org.osgi.annotation.versioning.ProviderType;
+
```
bjhargrave, line 21: This is another code move which the formatter will move back.

bosschaert: Yes you are right - my local experimentation project has different formatting rules. I'll line them up and fix the PR with them.

bosschaert: Fixed

bjhargrave: You should squash the commits into one commit.

stbischof:
There are still differen import Orders. In some files.

bjhargrave:
> There are still differen import Orders. In some files.

Yes, I see that too.

bosschaert:
> > There are still differen import Orders. In some files.
> 
> Yes, I see that too.

Doh sorry that should be fixed now.

PR #236, opened 2021-06-03
Change Features Class to Service Interface

bosschaert:
Now that Features is a Service spec, change the status Features class to
a Service Interface

### DIFF
--- a/org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
+++ b/org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- * SPDX-License-Identifier: Apache-2.0 
+ * SPDX-License-Identifier: Apache-2.0
  *******************************************************************************/
 package org.osgi.service.feature;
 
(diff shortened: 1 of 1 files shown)


org.osgi.service.feature/src/org/osgi/service/feature/Features.java
```diff
@@ -13,54 +13,41 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- * SPDX-License-Identifier: Apache-2.0 
+ * SPDX-License-Identifier: Apache-2.0
  *******************************************************************************/
 package org.osgi.service.feature;
 
+import org.osgi.annotation.versioning.ProviderType;
+
```
bjhargrave, line 21: Why is this moved? Are you using the Eclipse formatting configured for this project? We should not add these changes which formatting will change again. I am not saying you must use Eclipse to work on the code. But you still need to not change the code in ways that will change it again when formatted.

bosschaert: Fixed

bjhargrave: You should squash the commits into one commit.

stbischof:
An option would be to rename 'Features' into 'FeatureService'

bjhargrave:
> An option would be to rename 'Features' into 'FeatureService'

Yes, this makes a lot of sense.

bosschaert:
> > An option would be to rename 'Features' into 'FeatureService'
> 
> Yes, this makes a lot of sense.

I've made this change now.

bjhargrave (approved):

PR #238, opened 2021-06-06
[Converter] TypeReference has to be extended {}

stbischof:
Fix the doc examples, when using the Converter-TypeReference.
TypeReference  has to be extended.  `{}` was sometimes missing


bjhargrave (approved):

bjhargrave:
Thanks!

PR #245, opened 2021-06-11
Add support for Capabilities and Requirements on the Feature.

bosschaert:
Make it possible to explicitly specify Capabilities and Requirements on
the Feature. As described in spec section 8.


org.osgi.service.feature/src/org/osgi/service/feature/Feature.java
```diff
@@ -111,16 +120,20 @@
      */
     Map<String, FeatureExtension> getExtensions();
 
+	/**
+	 * Get the requirements as explicitly declared on the Features.
```
stbischof, line 124: Use singular `Features` -> `Feature`

bosschaert:
Closing this PR as this needs more thought. I would propose that we postpone requirements and capabilities to a future release of the Features.

PR #246, opened 2021-06-14
Run each TCK in its own GitHub Actions job

bjhargrave:
We use a build-time generated matrix to run each TCK in its own GitHub Actions job.

Some other minor changes are also in this PR.


rotty3000:
waah! super nice!

Issue #257, opened 2021-06-17
[ds] Changes to property and reference target precedence may break existing component descriptors


tjwatson:
The R7 section [112.6 Component Properties](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-component.properties) specifies the following:

> Each component configuration is associated with a set of component properties. The component properties are specified in the following configuration sources (in order of precedence):
> 1. Properties specified in the argument of the ComponentFactory.newInstance method. This is only applicable for factory components.
> 1. Properties retrieved from the OSGi Configuration Admin service in Configuration objects whose PID matches a configuration PID. The configuration PIDs are specified by the configuration-pid attribute of the component element. See Component Element. If no configuration-pid attribute is specified, the component name is used as the default configuration PID. If multiple configuration PIDs are specified, the order of precedence follows the order the configuration PIDs are specified in the component description. That is, the precedence for the configuration for an earlier specified configuration PID is lower than the precedence for the configurations for a later specified configuration PID.
> 1. Properties specified in the component description. Properties specified later in the component description override properties that have the same name specified earlier. Properties can be specified in the component description in the following ways:
>     * target attribute of reference elements - Sets the target property of the reference. See Target Property. The value of the target attribute is used for the value of a target property.
>     * property and properties elements - See Property and Properties Elements.

For R8 this section [112.6 Component Properties](https://osgi.github.io/osgi/cmpn/service.component.html#service.component-component.properties) has changed to the following:

> Each component configuration is associated with a set of component properties. The component properties are specified in the following configuration sources (in order of precedence):
> 1. Properties specified in the argument of the ComponentFactory.newInstance method. This is only applicable for factory components.
>1. Properties retrieved from the OSGi Configuration Admin service in Configuration objects whose PID matches a configuration PID. The configuration PIDs are specified by the configuration-pid attribute of the component element. See Component Element. If no configuration-pid attribute is specified, the component name is used as the default configuration PID. If multiple configuration PIDs are specified, the order of precedence follows the order the configuration PIDs are specified in the component description. That is, the precedence for the configuration for an earlier specified configuration PID is lower than the precedence for the configurations for a later specified configuration PID.
> 1. Properties specified in the component description. Properties specified later in the component description override properties that have the same name specified earlier. Properties can be specified in the component description in the following ways (in order of precedence):
>    1. property and properties elements - See Property and Properties Elements.
[...]

PR #260, opened 2021-06-17
Move explicit target property after explicit reference

tjwatson:
Related to #257

This updates the test to place the property for overriding the target of the explicit reference for the satisfying condition.  This should only be merged if we agree to the solution proposed in #257


tjwatson:
BJ and I discussed.   The DS 1.5 spec will be updated to have the component properties specified by the `property` or `properties` elements to take precedence over the property specified by the component reference target attributes, regardless of the order the `reference` element occurs in the component XML.

For DS 1.4 components the precedence order of the component properties remains the order specified in the component XML, regardless of the source element of the property (`reference`, `property`, `properties`).

PR #261, opened 2021-06-17
[Feature] validation of ID.class

stbischof:
We may should care more about validation of `ID.class`

- Constructor validation
- fix and validate fromMavenID method
- add java-doc and fix toString() method - use syntax
- remove usage of the word `Artifact ID` because ID could be member of Feature and Artifact.

---

### 
**After doing this I would like to suggest an alternative way:**
Provide just an Interface `ID` in the Feature-API. The Implementation then must handle the specified validation.
This may be done when adding additional methods:

**For Bundles:**
`BuilderFactory:newBundleBuilder(ID id)`

**additional:**
`BuilderFactory:newBundleBuilder(String groupId, String artifactId, String version)`
`BuilderFactory:newBundleBuilder(String groupId, String artifactId, String version, String type, String classifier)`
`BuilderFactory:newBundleBuilder(String mavenID)`

**For Features:**
`BuilderFactory:newFeatureBuilder(ID id)`

**additional:**
`BuilderFactory:newFeatureBuilder(String mavenID)`
`BuilderFactory:newFeatureBuilder(String groupId, String artifactId, String version)`
`BuilderFactory:newFeatureBuilder(String groupId, String artifactId, String version, String classifier)`

This would allow to grantee that the type of a if of the Feature must be `osgifeature`.

### DIFF
--- a/org.osgi.service.feature/src/org/osgi/service/feature/ID.java
+++ b/org.osgi.service.feature/src/org/osgi/service/feature/ID.java
@@ -17,16 +17,16 @@
  *******************************************************************************/
 package org.osgi.service.feature;
 
-import org.osgi.annotation.versioning.ProviderType;
-
 import java.util.Objects;
 import java.util.Optional;
 
+import org.osgi.annotation.versioning.ProviderType;
+
 /**
- * ID used to denote an artifact. This could be a feature model, a bundle which is part of the feature model
- * or some other artifact. <p>
- *
- * Artifact IDs follow the Maven convention of having:
+ * ID used to denote an artifact. This could be a feature model, a bundle which
+ * is part of the feature model or some other artifact.
+ * <p>
+ * IDs follow the Maven convention of having:
  * <ul>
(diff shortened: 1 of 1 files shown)


bosschaert (approved):

org.osgi.service.feature/src/org/osgi/service/feature/ID.java
```diff
@@ -151,8 +225,24 @@ public boolean equals(Object obj) {
...
+	 * @return the mavenID.
+	 */
     @Override
     public String toString() {
-        return groupId + ":" + artifactId + ":" + version;
+		StringBuilder sb = new StringBuilder(groupId).append(":")
+				.append(artifactId);
+
+		if (type != null) {
+			sb = sb.append(":").append(type);
```
stbischof, line 241: Remove sb=sb.app...

bjhargrave:
ID will become an interface.

PR #265, opened 2021-06-20
[Framework] ServiceEvent  add hint for UnfilteredServiceListener

stbischof:
ServiceEvents and BundleEvents must be immutable. 
So i think the ServiceProperties of the ServiceReference in a ServiceEvent should also be immutable. 
Same with the lastModified value of a Bundle in the BundleEvent.
I added a test to cover this. This test will fail.

I also added a hint about UnfilteredServiceListener to the ServiceListener.

### DIFF
--- a/org.osgi.framework/src/org/osgi/framework/ServiceEvent.java
+++ b/org.osgi.framework/src/org/osgi/framework/ServiceEvent.java
@@ -24,16 +24,17 @@ import java.util.EventObject;
 /**
  * An event from the Framework describing a service lifecycle change.
  * <p>
- * {@code ServiceEvent} objects are delivered to {@code ServiceListener}s and
- * {@code AllServiceListener}s when a change occurs in this service's lifecycle.
- * A type code is used to identify the event type for future extendability.
- * 
+ * {@code ServiceEvent} objects are delivered to {@code ServiceListener}s,
+ * {@code AllServiceListener}s and {@code UnfilteredServiceListener}s when a
+ * change occurs in this service's lifecycle. A type code is used to identify
+ * the event type for future extendability.
  * <p>
  * OSGi Working Group reserves the right to extend the set of types.
  * 
  * @Immutable
  * @see ServiceListener
  * @see AllServiceListener
+ * @see UnfilteredServiceListener
  * @author $Id$
  */
 


tjwatson:
I think the changes to the javadoc to give hints for UnfilteredServiceListener are good.

I have concerns about changing the behavior for event with respect to the `ServiceReference` and `Bundle` objects carried by events.  `FrameworkEvent`s also carry a `Bundle` object.  From the beginning of OSGi these objects contained in the events have been "live" objects that reflect the current state of the objects.  Changing them to no longer reflect the actual live state of the object will be a big change.  I would rather clarify the spec/javadoc to make it clear these objects are live.

Some complications to making `Bundle` immutable for events.  I assume the Bundle object associated with the ServiceReference must also be immutable.  What now happens to the methods on Bundle that change its state (e.g. start/stop/uninstall/update)?  Must these now throw some exception indicating the object is immutable?

I think other event mechanisms also would need to be considered, like LogEntry and the various `service` and `bundle` properties for event admin events.  If framework/service/bundle events use immutable types for bundle and servicereference then why would these other types events not?

stbischof:
In general i have no problem with a live Service reference.
My application does clone the the complete ServiceEvent.

But, then i found the @Immutable annotation in javadoc of Framework Bundle and ServiceEvent.

To i expected that the full object with all members is immutable.

So this would not be a new Feature wish. This is meaned as a test for the marked @Immutable javadoc

tjwatson:
Correct, the `@Immutable` javadoc on these event types is not adhered to by the current implementations for the bundle and service reference objects.  I argue that the original intent of the specification was not to make the `Bundle` nor `ServiceReference` objects included in these events to be immutable themselves.

I make this argument because the specification text does not give all the necessary details that would be required for an implementation to make such a guarantee.  To make such a guaranteed immutable event there are many things that would need to be worked out.  For example, all the various objects which are returned by `Bundle.adapt`, these would need to be immutable.  For installed bundles, the `Bundle.getEntry` methods would have to keep returning valid URLs even if the backing bundle has been uninstalled and removed from disk after a refresh operation.  Once a bundle is resolved it may have fragments attached, this influences things like Bundle.findEntries.  But if a listener stashed the immutable `Bundle` object from an INSTALLED event then this Bundle.findEntries must act as if the bundle is not resolved.  For an active `Bundle` the `Bundle.getBundleContext` method would have to continually return a valid BundleContext, even if the "real" bundle has been stopped or uninstalled.  This is only a few examples, more dragons can be found in providing access to an immutable bundle class loader.

I think the best thing here is to make it clear these `Bundle` and `ServiceReference` objects are live.  If a snapshot is needed then these objects should be adapted to their respective DTOs by the listeners receiving the events.

bjhargrave:
It is never the case that an immutable object must be deeply immutable. That is, an immutable object can reference mutable object. Otherwise nothing would really work. :-)

The ServiceEvent, BundleEvent, and FrameworkEvent objects are in fact immutable in that all their fields are final. They do however reference mutable objects.

We must not change ServiceEvent, BundleEvent, and FrameworkEvent to attempt to make the references object also immutable.

bjhargrave (approved):

stbischof:
@bjhargrave  I removed the most.

I just saw the `@Immutable` and after reading the following. I not expectet that the ServiceReference will change. But i am fine with this.
https://docs.oracle.com/javase/tutorial/essential/concurrency/imstrat.html
```
A Strategy for Defining Immutable Objects
...
4. If the instance fields include references to mutable objects, don't allow those objects to be changed:
   - Don't provide methods that modify the mutable objects.
   - Don't share references to the mutable objects. Never store references to external, mutable objects passed to the constructor; if necessary, create copies, and store references to the copies. Similarly, create copies of your internal mutable objects when necessary to avoid returning the originals in your methods.
```

PR #269, opened 2021-06-22
ds: Clarify spec regarding precedence of target properties

bjhargrave:
Closes https://github.com/osgi/osgi/issues/257


tjwatson (approved):
LGTM

PR #271, opened 2021-06-24
TypedEvent TCK - initial

stbischof:
Draft- To see what the CI will say.


timothyjward:
> i could run it local but i cant find out why this errors in the build came up.

Did you add the typed event implementation to the [list of compendium implementations](https://github.com/osgi/osgi/blob/b5e212ddf80748a68ce7773b4deaba37b6c219eb/osgi.impl/layout.bnd#L36-L37)? This is what causes the implementation to be packaged up and run with the CT in the build. Without it the CT will fail due to a missing implementation.

bjhargrave:
> i could run it local

By local, I assume you mean the `testOSGi` task. This is useful to you as a developer but it is not how the TCK will be used by someone else who will not have this repo build on their system. The CI build builds the tck jars which are passed to the TCK execution jobs to run the TCKs as they are intended to run. See the CI build for how this is done.

stbischof:
Thank you @timothyjward! This solved my problem.

cnf/repo/org.osgi.impl.service.typedevent/org.osgi.impl.service.typedevent-8.0.0.lib
```diff
@@ -0,0 +1,8 @@
+org.osgi.service.typedevent; version=latest
+org.osgi.impl.service.component; version=latest
+org.apache.aries.typedevent.bus; version=0.0.1
+org.osgi.util.pushstream; version=latest
+org.apache.aries.component-dsl.component-dsl; version=latest
```
bjhargrave, line 5: This should be a specific version. 1.2.2 in this case.

org.osgi.test.cases.typedevent/bnd.bnd
```diff
@@ -0,0 +1,23 @@
...
+
+-conditionalpackage					= org.osgi.test.support.*
+-privatepackage						=  \
+	${p}.*
+Import-Package: ${-signaturetest}, *
+
+-buildpath: \
+	org.osgi.test.support;version=project,\
+	org.osgi.service.typedevent;version=latest,\
+	org.osgi.framework,\
```
bjhargrave, line 12: For both framework and dto, you need to specify an actual version number and also `maven-scope=provided` since they come from the framework at runtime.

```
    org.osgi.framework;maven-scope=provided;version=1.8,\
    org.osgi.dto;maven-scope=provided;version=1.0
```

org.osgi.test.cases.typedevent/bnd.bnd
```diff
@@ -0,0 +1,23 @@
...
+	org.osgi.dto
+
+-runbundles						= \
+    org.osgi.impl.service.log;version=latest, \
+    org.osgi.impl.service.cm;version=latest, \
+	org.osgi.service.typedevent;version=latest, \
```
bjhargrave, line 18: org.osgi.impl.service.typedevent already include this so it does not need to be re-included.

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TestEvent.java
```diff
@@ -0,0 +1,16 @@
+package org.osgi.test.cases.typedevent.junit;
```
bjhargrave, line 1: All java files need to start with the license notice comment.

bjhargrave: This is not a junit test class, so it should not be in the junit package which is generally only the junit tests. So something like `org.osgi.test.cases.typedevent.dto`.

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TestEvent.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.osgi.dto.DTO;
+
+public class TestEvent extends DTO {
+
+	public TestEvent(String test, String a, String A) {
```
bjhargrave, line 7: DTOs have no behavior which means no constructors. While the spec supports DTO-like object, this class extends DTO but violates the DTO spec.

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TestEventHandler.java
```diff
@@ -0,0 +1,57 @@
+package org.osgi.test.cases.typedevent.junit;
```
bjhargrave, line 1: license notice missing and package name

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventBusTest.java
```diff
@@ -0,0 +1,169 @@
...
+@ExtendWith(ServiceExtension.class)
+@Nested
+public class TypedEventBusTest {
+
+	private static final DTO	dto	= new DTO() {
+										public String a = "";
+									};
+
+	@InjectService
+	static TypedEventBus		typedEventBus;
```
bjhargrave, line 48: Lets avoid statics. It does not cost much to inject on each test method.

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventHandlerTest.java
```diff
@@ -0,0 +1,157 @@
...
+	class B extends A {
+		String bb;
+	}
+
+	class C extends B {
+		String cc;
+	}
+
+	@InjectService
+	static TypedEventBus bus;
```
bjhargrave, line 57: Lets avoid statics. It does not cost much to inject on each test method.

bjhargrave (changes requested):

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventBusTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+		assertThatNullPointerException().isThrownBy(() -> {
+			typedEventBus.deliverUntyped("a", null);
+		});
+	}
+
+	@ParameterizedTest
+	@ValueSource(strings = {
+			"*", "%", "/", "//", "a//b", "//b", "a//", "a/", "a/*/*/b",
+			"a/b/c/", "a/*/*", " /*", " ", "a/ /b", " / ", "Bär"
```
timothyjward, line 113: This is a nicely complete test, but I'm pretty sure that `Bär` is a valid topic. To the best of my knowledge `ä` is a valid `jLetterOrDigit` and so the topic token should be permitted by the implementation.

timothyjward:
I've made an update to the RI which fixes all but one of the failing tests. I believe that the one remaining failure is a bug in the TCK, and have commented on it in this PR.

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventBusTest.java
stbischof: `Character.isJavaIdentifierPart('ä')=true` it is. i will fix that

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventBusTest.java
```diff
@@ -0,0 +1,184 @@
...
+			"1", "2/3", "4/a/5", "a/_", "-/-", "_/_", "-", "_", "$",
+			"$/$1$a$/$", "Bär"
+	})
+	public void test_deliver_2_topic_valid_should_not_throws_exception(
+			String validTopic) {
+		typedEventBus.deliver(validTopic, dto);
+	}
+
+	@Nested
+	static class NotSpecified_ {
```
stbischof, line 144: Remove this 3 tests

stbischof:
@bjhargrave @timothyjward 
The TCK is now running.
Will add additional Tests next Days especially on Event Mutability and Thread Safety

PR #272, opened 2021-06-24
Turn ID into an interface

bosschaert:

### DIFF
--- a/org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
+++ b/org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
@@ -33,7 +33,53 @@ public interface FeatureService {
      * Get a factory which can be used to build feature model entities.
      * @return A builder factory.
      */
-    public BuilderFactory getBuilderFactory();
+	BuilderFactory getBuilderFactory();
+
+	/**
+	 * Obtain an ID from a Maven Coordinates formatted string. The supported
+	 * syntax is as follows:
+	 * <p>
+	 * {@code groupId ':' artifactId ( ':' type ( ':' classifier )? )? ':' version }
+	 * 
+	 * @param coordinates The Maven Coordinates.
+	 * @return the ID.
+	 */
+	ID getIDfromMavenCoordinates(String coordinates);
+
+	/**
+	 * Obtain an ID.
+	 * 
+	 * @param groupId The group ID (not {@code null}, not empty).
+	 * @param artifactId The artifact ID (not {@code null}, not empty).
+	 * @param version The version (not {@code null}, not empty).
+	 * @return The ID.
+	 */
+	ID getID(String groupId, String artifactId, String version);
+
+	/**
+	 * Obtain an ID.
+	 * 
+	 * @param groupId The group ID (not {@code null}, not empty).
+	 * @param artifactId The artifact ID (not {@code null}, not empty).
+	 * @param version The version (not {@code null}, not empty).
+	 * @param type The type (not {@code null}, not empty).
+	 * @return The ID.
+	 */
+	ID getID(String groupId, String artifactId, String version,
+			String type);
+
+	/**
+	 * Obtain an ID.
+	 * 
+	 * @param groupId The group ID (not {@code null}, not empty).
+	 * @param artifactId The artifact ID (not {@code null}, not empty).
+	 * @param version The version (not {@code null}, not empty).
+	 * @param type The type (not {@code null}, not empty).
+	 * @param classifier The classifier (not {@code null}, not empty).
+	 * @return The ID.
+	 */
+	ID getID(String groupId, String artifactId, String version,
+			String type, String classifier);
 
     /**
      * Read a Feature from JSON
@@ -41,7 +87,7 @@ public interface FeatureService {
      * @return The Feature represented by the JSON
      * @throws IOException When reading fails
      */
-    public Feature readFeature(Reader jsonReader) throws IOException;
+	Feature readFeature(Reader jsonReader) throws IOException;
 
     /**
      * Write a Feature Model to JSON
@@ -49,5 +95,5 @@ public interface FeatureService {
      * @param jsonWriter A Writer to which the Feature should be written.
      * @throws IOException When writing fails.
      */
-    public void writeFeature(Feature feature, Writer jsonWriter) throws IOException;
+	void writeFeature(Feature feature, Writer jsonWriter) throws IOException;
 }
--- a/org.osgi.service.feature/src/org/osgi/service/feature/ID.java
+++ b/org.osgi.service.feature/src/org/osgi/service/feature/ID.java
@@ -17,15 +17,14 @@
  *******************************************************************************/
 package org.osgi.service.feature;
 
-import org.osgi.annotation.versioning.ProviderType;
-
-import java.util.Objects;
 import java.util.Optional;
 
+import org.osgi.annotation.versioning.ProviderType;
+
 /**
- * ID used to denote an artifact. This could be a feature model, a bundle which is part of the feature model
- * or some other artifact. <p>
- *
+ * ID used to denote an artifact. This could be a feature model, a bundle which
+ * is part of the feature model or some other artifact.
+ * <p>
  * Artifact IDs follow the Maven convention of having:
  * <ul>
  * <li>A group ID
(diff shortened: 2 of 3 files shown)


stbischof:
maybe we can add a ::toString method to add a javaDoc that specifies that a Sting with the GAV-Syntax is returned.

bosschaert:
> maybe we can add a ::toString method to add a javaDoc that specifies that a Sting with the GAV-Syntax is returned.

Good point - I just added that.

org.osgi.service.feature/src/org/osgi/service/feature/ID.java
```diff
@@ -34,125 +33,53 @@
...
+	Optional<String> getClassifier();
 
-    @Override
-    public String toString() {
-        return groupId + ":" + artifactId + ":" + version;
-    }
+	/**
+	 * This method returns the ID using the following syntax:
+	 * <p>
+	 * {@code groupId ':' artifactId [ ':' type [ ':' classifier ] ] ':' version }
```
bjhargrave, line 79: Like this doc. But you should use the syntax rules from [Core 1.3.2](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.introduction.html#framework.general.syntax).

`groupId ':' artifactId ( ':' type ( ':' classifier )? )? ':' version`

`[...]` means chose one from a set

org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
```diff
@@ -35,6 +35,49 @@
...
+	 * Obtain an ID.
+	 * 
+	 * @param groupId The group ID.
+	 * @param artifactId The artifact ID.
+	 * @param version The version.
+	 * @param type The type.
+	 * @param classifier The classifier.
+	 * @return The ID.
+	 */
+	public ID getID(String groupId, String artifactId, String version,
```
stbischof, line 78: Type must not be null is classifier is not null.

bosschaert: Hi @stbischof there are 3 overloads for `getID()` - none of them should be used with `null` arguments but they express what is allowable through the method signature.

I also dded a constant for the Feature Type into the ID interface. I'm not sure an extra method to create IDs for features adds much value if we already have the constant.

stbischof: that is enough.

stbischof: last edge-case:
- What happens with empty values or if they contain a ":"?

bjhargrave: For each argument, the javadoc should state whether the value must not be null or may be null. Then it is clear to the reader and the TCK implementor.

bosschaert: > last edge-case:
> 
> * What happens with empty values or if they contain a ":"?

Would that not be outside the grammar definition?

    groupId ':' artifactId ( ':' type ( ':' classifier )? )? ':' version

which says: a ':' is _always_ followed by a value, or am I misunderstanding?

bjhargrave: I guess it should be clear that the productions for `groupId` etc. result in non-empty tokens.

org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
```diff
@@ -33,21 +33,67 @@
...
+	 */
+	ID getID(String groupId, String artifactId, String version);
+
+	/**
+	 * Obtain an ID.
+	 * 
+	 * @param groupId The group ID (not {@code null}).
+	 * @param artifactId The artifact ID (not {@code null}).
+	 * @param version The version (not {@code null}).
+	 * @param type The type (or {@code null}).
```
bjhargrave, line 65: If null, default is? `jar`?

org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
```diff
@@ -33,21 +33,67 @@
...
+			String type);
+
+	/**
+	 * Obtain an ID.
+	 * 
+	 * @param groupId The group ID (not {@code null}).
+	 * @param artifactId The artifact ID (not {@code null}).
+	 * @param version The version (not {@code null}).
+	 * @param type The type (or {@code null}).
+	 * @param classifier The classifier (or {@code null}).
```
bjhargrave, line 78: If null, default is? `` (empty string)?

If non-null, doesn't type need to be non-null? Or does it default to `jar`?

org.osgi.service.feature/src/org/osgi/service/feature/ID.java
```diff
@@ -34,125 +33,58 @@
...
-    @Override
-    public int hashCode() {
-        return Objects.hash(artifactId, classifier, groupId, type, version);
-    }
+	/**
+	 * Get the type identifier.
+	 * 
+	 * @return The type identifier.
+	 */
+	Optional<String> getType();
```
bjhargrave, line 72: I am now thinking this should not be optional but instead be just String where the default value is `jar`

org.osgi.service.feature/src/org/osgi/service/feature/ID.java
```diff
@@ -34,125 +33,58 @@
...
-        return Objects.equals(artifactId, other.artifactId) && Objects.equals(classifier, other.classifier)
-                && Objects.equals(groupId, other.groupId) && Objects.equals(type, other.type)
-                && Objects.equals(version, other.version);
-    }
+	/**
+	 * Get the classifier.
+	 * 
+	 * @return The classifier.
+	 */
+	Optional<String> getClassifier();
```
bjhargrave, line 79: I am now thinking this should not be optional but instead be just String where the default value is `` (empty string)

These changes are necessary since if classifier has a value, type must also have a value. So better to just make sure they always have actual values.

org.osgi.service.feature/src/org/osgi/service/feature/ID.java
```diff
@@ -34,125 +33,58 @@
...
 
-    @Override
-    public String toString() {
-        return groupId + ":" + artifactId + ":" + version;
-    }
+	/**
+	 * This method returns the ID using the following syntax:
+	 * <p>
+	 * {@code groupId ':' artifactId ( ':' type ( ':' classifier )? )? ':' version }
```
bjhargrave, line 84: To be more clear here:

- classifier is included if its value is not `` (empty string).
- type is included if its value is not `jar` or classifier's value is not `` (empty string).

org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
bosschaert: Default is that this element isn't there. I.e. there is no default.

org.osgi.service.feature/src/org/osgi/service/feature/ID.java
bosschaert: No, if the type is not specified there is no type. In some contexts 'jar' is then used instead, but from an ID point of view, if it's not specified then there is no value.

org.osgi.service.feature/src/org/osgi/service/feature/ID.java
bosschaert: I think an alternative would be to use `null` if there is no value, but I would _not_ use an empty string.

bosschaert:
I think it's actually cleaner to simply not allow `null` to any of the ID factory methods. That way you can never have a classifier where the type is not specified. 
I would not infer any default values of type or classifier. They are either there or they are not. When the classifier is there, the type is always also there. The API expresses this.
I have pushed a small update to represent this.

org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
bjhargrave: So empty Optional and Optional of empty string mean the same thing for classifier? And empty Optional and Optional of `jar` mean the same thing for type? Otherwise how is toString to properly render the string? You don't want `::` when classifier is an Optional of empty string.

org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
bjhargrave: As above, this is not really true.

bjhargrave:
> That way you can never have a classifier where the type is not specified.

It also means that you can never specify a type and not specify a classifier. Which means that for classifier empty Optional and Optional of empty string must be treated identically.

bjhargrave:
You may want to look at the api Bnd has for this: https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/maven/api/Archive.java.

stbischof:
Does that mean that the ID of a Feature itsselfe always needs to have a type set?

when using it like in the spec examples
```
{
    "id": "org.acme:acmeapp:2.0.0",
      "name": "The Acme Application"
}
```
it should load a jar?

bjhargrave:
> Does that mean that the ID of a Feature itsselfe always needs to have a type set?

I think type must always default to `jar` if not specified. This is the maven way. GAV alone does not provide enough information about the artifact since a GAV generally has multiple artifacts (like jar, pom).

bosschaert:
> > That way you can never have a classifier where the type is not specified.
> 
> It also means that you can never specify a type and not specify a classifier. Which means that for classifier empty Optional and Optional of empty string must be treated identically.

No, there is a factory method that has just a type and not a classifier: https://github.com/osgi/osgi/pull/272/files#diff-0cf53bfe27354ab0be03ceead60dfc5032ecbb70bc3f44483992756309eccca4R68

bosschaert:
> > Does that mean that the ID of a Feature itsselfe always needs to have a type set?
> 
> I think type must always default to `jar` if not specified. This is the maven way. GAV alone does not provide enough information about the artifact since a GAV generally has multiple artifacts (like jar, pom).

If the type is not specified, the default for Maven is context-dependent. Maybe in the context of Bnd it can always be evaluated to `jar` but for example in the context of a pom the default is `pom`, e.g. see here for an example where a parent pom is referred to without explicitly declaring the type: https://github.com/apache/felix-dev/blob/master/framework/pom.xml#L22

The ID interface in features provides a general mechanism to specify an entity's ID. We should follow the Maven way of using the context to infer what the default is if it's not specified. In the context of bundles it will be `jar` but in another context it may be something else.

So, if the type is not specified when obtaining the ID, the ID should simply not contain a value for the type so it can be inferred in a context-sensitive manner when needed.

bjhargrave:
> he default for Maven is context-dependent.

How many contexts are their in the Feature spec? Shouldn't these be specified so that the default for type is known to all?

If you want to specify a classifier, you must specify a type. So there must be a well known default type for people to use in this case.

If I do `getID("g", "a", "1", "jar", "")` what will `toString()` return? Will it be different than `getID("g", "a", "1")`?

Part of that question is whether empty string is a valid classifier?

stbischof:
At the moment we use the ID for the bundles. There it will be a jar.
But we also have an id in the id attribute of the Feature it's self. Then it would be osgifeature.

Also if we will have referenced Features  in the Feature to comvine then then there may will be an additional features reference beside the bundles. Ther also would be osgifeature the default type

bosschaert:
> How many contexts are their in the Feature spec? Shouldn't these be specified so that the default for type is known to all?

Right now there are 2 main contexts: IDs specified for bundles and the ID specified for the feature itself. In the case of bundle the default should be `jar` and in the case of the feature the default should be `osgifeature`. I think it would be good to document this in the spec (currently it isn't).

However there could be additional different defaults for IDs used in extensions, what the default here is should be part of the extension's documentation.

> If you want to specify a classifier, you must specify a type. So there must be a well known default type for people to use in this case.

Agreed.

> If I do `getID("g", "a", "1", "jar", "")` what will `toString()` return? Will it be different than `getID("g", "a", "1")`?

    getID("g", "a", "1", "jar", "")
      -> I think this one should throw an IllegalArgument exception, an empty string classifier should not be allowed

And yes, the following should have different output IMHO:

    getID("g", "a", "1", "jar").toString() -> "g:a:jar:1"
    getID("g", "a", "1").toString() -> "g:a:1"


> Part of that question is whether empty string is a valid classifier?

I don't think an empty string is a valid argument for any of: groupID, artifactID, version, type, classifier. We should add this to the javadoc.

bosschaert:
I have squashed all the commits in this PR into one and updated the Javadoc with default interpretations of `type` for feature and bundle. Also I marked all factory methods for ID _not_ to allow empty values which also implies non-null.

bjhargrave:
> Also I marked all factory methods for ID _not_ to allow empty values which also implies non-null.

`(must not be empty)` does not also mean must not be null in that "null" is not a subset of "empty".

bosschaert:
> > Also I marked all factory methods for ID _not_ to allow empty values which also implies non-null.
> 
> `(must not be empty)` does not also mean must not be null in that "null" is not a subset of "empty".

I assumed that non-empty also means not null and not `""`, but I can add `not null` explicitly if that helps.

bosschaert:
I changed the Javadoc from `(must not be empty)` to `(not null, not empty)` and squashed everything in my latest commit.

bjhargrave (approved):

org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
```diff
@@ -33,21 +33,67 @@
...
+	/**
+	 * Obtain an ID from a Maven ID formatted string. The supported syntax is as
+	 * follows:
+	 * <p>
+	 * {@code groupId ':' artifactId ( ':' type ( ':' classifier )? )? ':' version }
+	 * 
+	 * @param mavenID The Maven ID.
+	 * @return the ID.
+	 */
+	ID getIDfromMavenID(String mavenID);
```
stbischof, line 47: Is mavenId a common word? I only used Maven Coordinates in the past.

bosschaert: You're right. I always called these 'Maven IDs' but looking at https://maven.apache.org/pom.html#Maven_Coordinates coordinates seems the right term. I changed the parameter name in the latest squashed commit.

org.osgi.service.feature/src/org/osgi/service/feature/FeatureService.java
```diff
@@ -33,21 +33,67 @@
...
+	/**
+	 * Obtain an ID from a Maven Coordinates formatted string. The supported
+	 * syntax is as follows:
+	 * <p>
+	 * {@code groupId ':' artifactId ( ':' type ( ':' classifier )? )? ':' version }
+	 * 
+	 * @param coordinates The Maven Coordinates.
+	 * @return the ID.
+	 */
+	ID getIDfromMavenID(String coordinates);
```
bjhargrave, line 47: The method name is still using "MavenID". I assumed @stbischof was commenting on the method name not just the parameter name. This method should be another overload  of `getID`.

bosschaert: Sorry guys, clearly missed that one. I just fixed the method name (now `getIDfromMavenCoordinates()`), squashed and pushed again.

stbischof:
LGTM

PR #274, opened 2021-06-24
[Features Service] Remove section about Capabilities and Requirements

bosschaert:


stbischof:
I agree to remove this for the moment. Maybe we have a comeback when discussion about merging ans launching features.

PR #281, opened 2021-07-02
Use a FeatureArtifact to build extensions of type artifact

bosschaert:
This is consistent with the return of objects of type FeatureArtifact
already done by FeatureExtension.getArtifacts()


org.osgi.service.feature/src/org/osgi/service/feature/FeatureExtensionBuilder.java
```diff
@@ -48,34 +48,10 @@
 	 * Add an Artifact to the extension. Can only be called for extensions of
 	 * type {@link FeatureExtension.Type#ARTIFACTS}.
 	 *
-	 * @param aid The ArtifactID of the artifact to add.
+	 * @param art The artifact to add.
```
bjhargrave, line 51: We can't afford to use the whole word `artifact` for the parameter name? :-)

Since this is a specification, we should avoid using unnecessarily abbreviated parameter names.

bosschaert: Good point - I've addressed this here and in a few other places too.

Issue #266, opened 2021-06-21
[Feature] Handling of feature ID's and Json


maho7791:
This discussion refers to Pull-Request: https://github.com/osgi/osgi/pull/261

For me the features ID, that is a GAV String, similar to what we do with `org.osgi.framework.Filter`. We have the ' FrameworkUtil` that is the factory for creating Filter instances, where Filter is an Interface. This factory provides the validation of the provided filter in respect to the syntax. So it takes care that the provided filter string is correct.
Beside this, we use the filter-strings in our service properties, that are later converted into Filter instances.

For the feature ID that would mean, making it an interface. We have to provide one or more factory methods, with corresponding signatures (maybe from the current class implementation), that validate the provided parameters and ensure that we get a valid ID instance. In this case it would be possible to delegate the implementation of the validation to specification implementors. The specification should then define the validation rules.

That would also make it possible to define the type of an ID in e.g. the `org.osgi.service.feature.FeatureArtifact` as a String to must be a valid ID-String. That would make the corresponding JSON, more compliant to common Json-Serializers. So a Json for a Feature artifact could always look like this:
```
{
  "id": "org.acme:acmeapp:1.0.1",
  "name": "The Acme Application",
  "license": "https://opensource.org/licenses/Apache-2.0",
  "isComplete": true,
  "bundles": [
        { "id": "org.osgi:org.osgi.util.function:1.1.0"},
        { "id": "org.osgi:org.osgi.util.promise:1.1.1"},
        {
            "id": "org.apache.commons:commons-email:1.5",
            "metadata":[
                 {"org.acme.javadoc.link": "https://commons.apache.org/proper/commons-email/javadocs/api-1.5"},
                 {"id": "test"}
             ]
      },
      { "id": "com.acme:acmelib:1.7.2"}      
    ]
}
```
instead of the format in the current. Because especially when using the metadata, this format would allow, having a property with key id in the metadata map. When having the valid Json like this:
```
{
  "id": "org.acme:acmeapp:1.0.1",
  "name": "The Acme Application",
  "license": "https://opensource.org/licenses/Apache-2.0",
  "isComplete": true,
  "bundles": [
      "id": "org.osgi:org.osgi.util.function:1.1.0",
      "id": "org.osgi:org.osgi.util.promise:1.1.1",
      {
          "id": "org.apache.commons:commons-email:1.5",
          "org.acme.javadoc.link": "https://commons.apache.org/proper/commons-email/javadocs/api-1.5",
[...]

PR #294, opened 2021-08-06
Changes to the Features Spec

### LINKED ISSUE #206: [Features] Rename Feature-JSON field `isComplete`
the Field `isComplete` looks like an getter/method call.
We may should change it to `complete`

bosschaert:
* Rename 'isComplete' to 'complete' attribute
* Remove support for shorthand way to specify bundles and artifacts
* Clarify use of reverse DNS naming on custom metadata

Fixes #206 and #266

### DIFF
--- a/osgi.specs/docbook/159/service.feature.xml
+++ b/osgi.specs/docbook/159/service.feature.xml
@@ -246,6 +246,13 @@ Features may have dependencies on other Features. Features inherit the capabilit
             </row>
 
             <row>
+              <entry>complete</entry>
+              <entry>boolean</entry>
+              <entry>Optional, defaults to <code>false</code></entry>
+              <entry>Completeness of the Feature. A Feature is complete when it has no external dependencies.</entry>
+            </row>
+
+            <row>
               <entry>description</entry>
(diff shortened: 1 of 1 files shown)


stbischof:
As I remember we would avoid  the `id` clash between the artefacts id and an id as an property with en attrebute propertieds that holds the map , not just with an doc about dns. 

So it does not fix  #266

@maho7791 Right?

bosschaert:
> As I remember we would avoid the `id` clash between the artifacts id and an id as an property with en attribute properties that holds the map , not just with an doc about dns.
> 
> So it does not fix #266
> 
> @maho7791 Right?

Hi @stbischof please see this comment regarding what was discussed in the Spec Project call: https://github.com/osgi/osgi/issues/266#issuecomment-892779098

maho7791:
After talking with David, I agree the changes.

bosschaert:
Thanks @maho7791 - we discussed some additions such as a metamodel which can be added in the future for R9.

@stbischof if you're ok too, I'll merge this change.

stbischof:
agree!

PR #295, opened 2021-08-06
Add Feature Resource Versioning to spec chapter

### LINKED ISSUE #209: [Feature] Feature-Json should provide resource-version and JSON Schema
A Global property in the Feature-Json specifying the version of the resource format might also be usefull. We have this in
- Configurator: `":configurator:resource-version" : 1,`
- Manifest: `Manifest-Version: 1.0`

From @bosschaert: 
> We could also use a JSON Schema and encode the version in the schema URL, similar to how this is 
> done for XML Schemas in the Remote Service Admin spec https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.remoteserviceadmin.html#d0e92612

extracted from #179

bosschaert:
This fixes #209 (the JSON schema was postponed)

### DIFF
--- a/osgi.specs/docbook/159/service.feature.xml
+++ b/osgi.specs/docbook/159/service.feature.xml
@@ -711,7 +711,6 @@ Feature f = builder.build();</programlisting>
     </section>
   </section>
 
-
   <section>
     <title>Framework Launching Properties</title>
     <para>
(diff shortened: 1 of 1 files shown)


osgi.specs/docbook/159/service.feature.xml
```diff
@@ -729,6 +729,29 @@ Feature f = builder.build();</programlisting>
...
+  "id": "org.acme:acmeapp:1.0.0"
+
+  /*
+    Additional Feature entities here
+    ...
+  */
+}</programlisting>
+    <para>
+    The currently supported version of the Feature JSON documents is 1.0. If no
+    Feature Resource Version is specified <code>1.0</code> is used as the default.
```
bjhargrave, line 752: For the example JSON documents in the spec, we should probably make sure they all include `"feature-resource-version": "1.0"` to encourage others to do so.

bosschaert: I agree. However I would like to do this in a separate PR. There are a number of PRs open right now already and making that change would probably cause merge conflicts...

bosschaert: As I had to rebase anyway I have now added the `"feature-resource-version"` to all the JSON snippets.

PR #300, opened 2021-08-10
Small editorial change to the Feature Extension spec text.

bosschaert:

### DIFF
--- a/osgi.specs/docbook/159/service.feature.xml
+++ b/osgi.specs/docbook/159/service.feature.xml
@@ -529,7 +529,7 @@ Feature f = builder.build();</programlisting>
   <section>
     <title>Extensions</title>
     <para>
-      Features can be extended with custom content. This makes it possible to keep custom entities and information relating to the Feature
+      Features can include custom content. This makes it possible to keep custom entities and information relating to the Feature
     together with the rest of the Feature.</para>
     
     <para>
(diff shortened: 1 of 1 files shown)


stbischof:
how could it be handled to use the same extension multiple times with different configurations?

like with the `config.factory.pattern~here`?

bosschaert:
Hi @stbischof it should depend on the actual extension whether multiple instances of it make sense.

I can see that it would make sense if you for example have an extension that defines validation rules for configuration to have multiple instances of those. However the extension mechanism itself already supports multiple values (in this case you could use a JSON extension that is an array of JSON objects, one for each validation rule), so registering the extension itself multiple times would not be needed.

On the other hand having multiple instances doesn't make sense for some other extensions. For example an extension that specifies the framework launcher properties to be set only makes sense to be there once.

An extension already defines how it interprets its provided definition. By letting the extension itself define whether it can handle multiple values in its content you basically already ensure that only multiple values can be specified for the extension when it makes sense, and you won't end up with situations where multiple values are specified in cases where it doesn't make sense.
At least this is my experience.

On the other hand if there is a strong use-case for this, we could quite easily add this in a future version. Given that we are so close to the deadline I would suggest not to add this new concept right now.

stbischof:
Fully agree to not handle it in this release. Just want so speak it out to be sure we have the options to do more later.

PR #317, opened 2021-08-30
Using Temurin builds of OpenJDK in GH actions

carldea:
The AdoptOpenJDK has been discontinued Since July 2021. When using Zulu you get all the latest updated TCK tested builds for all versions of OpenJDK.

### DIFF
--- a/.github/workflows/bnd-test.yml
+++ b/.github/workflows/bnd-test.yml
@@ -47,7 +47,7 @@ jobs:
     - name: Set up Java
       uses: actions/setup-java@v2
       with:
-        distribution: 'adopt'
+        distribution: 'temurin'
         java-version: ${{ env.java_version }}
     - name: Build
       id: build
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -64,7 +64,7 @@ jobs:
     - name: Set up Java
       uses: actions/setup-java@v2
       with:
-        distribution: 'adopt'
+        distribution: 'temurin'
         java-version: ${{ env.java_version }}
     - name: Build
       id: build
@@ -105,7 +105,7 @@ jobs:
     - name: Set up Java
       uses: actions/setup-java@v2
       with:
-        distribution: 'adopt'
+        distribution: 'temurin'
         java-version: ${{ env.java_version }}
     - name: Build
       id: build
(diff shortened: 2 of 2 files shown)


bjhargrave:
Why would we not replace `adopt` with `temurin`?

carldea:
> Why would we not replace `adopt` with `temurin`?

I do get that question often. While Temurin is a fine choice. Azul is just consistent supporting every version of OpenJDK even JDK 17. Temurin only supports fewer versions. Of course both distros  are TCK tested. Azul was the default choice for setup-java@v1.  

Many commercial enterprise environments in production are moving to Zulu because of its maturity (fast security updates). 

If you would prefer, I can change the distro. Or create a matrix (an array) of distros, to build artifacts on both distros.

bjhargrave:
I would prefer to use the one which is already in the OS image. That is currently `adopt` and I assume will be changing to `temurin`: https://github.com/actions/virtual-environments/issues/3859. So I would prefer `temurin`.  I don't see the need for a matrix to include `zulu`.

Also, you need to amend your commits so that each commit message includes a `Signed-off-by` as required by CONTRIBUTING. Thanks.

carldea:
> I would prefer to use the one which is already in the OS image. That is currently `adopt` and I assume will be changing to `temurin`: [actions/virtual-environments#3859](https://github.com/actions/virtual-environments/issues/3859). So I would prefer `temurin`. I don't see the need for a matrix to include `zulu`.
> 
> Also, you need to amend your commits so that each commit message includes a `Signed-off-by` as required by CONTRIBUTING. Thanks.

I committed the following:
- Changed distro to Temurin
- Included Signed-off 

Respectfully,
Carl

bjhargrave:
I just noticed the bnd-test workflow also needs the same update.

carldea:
> I just noticed the bnd-test workflow also needs the same update.

I believe you meant `cibuild.yml`. 
Will do.

bjhargrave:
> I believe you meant `cibuild.yml`.

Yes. There are 2 workflows which need the change.

carldea:
The change is complete. However, I believe one of the tests failed, two tests are still running.
https://github.com/carldea/osgi/runs/3467257052?check_suite_focus=true

Not sure if a forked work flow (mine) needs access to remote resources to run the TCK test properly.

bjhargrave:
> However, I believe there are two tests that failed.

Yes, we have 2 TCK runs which have known failures for which we have open issues. 

Thanks for the PR! I kicked off the PR build and will merge when complete.

Issue #322, opened 2021-09-07
[Typed Event] TCK Missing coverage areas


timothyjward:
- [x] No test DTOs have nested data (157.2.1.1) 
- [x] No checking of nested data in untyped DTOs (157.2.2) 
- [x] No use of auto-generated topics in delivery (157.3)
- [x] No use of conversion to reified type in a TypedEventHandler (157.4.1)
- [x] No use of service properties to override reified types in a TypedEventHandler (157.4.1)
- [x] No use of event filters (157.4.5) 
- [x] No tests of handler error states (157.4.8) 
- [x] No tests for the Event Monitor service (157.6) 
- [x] No check for the spec mandated capabilities (157.7)

stbischof:
I have a view on this and some more tests that uses the pattern like your tests.

https://github.com/apache/aries-typedevent/blob/main/org.apache.aries.typedevent.bus/src/test/java/org/apache/aries/typedevent/bus/osgi/EventDeliveryIntegrationTest.java#L108

Waiting to get Mockito in with the TCK for the features.

timothyjward:
The Typed Events TCK now has good coverage of the specification

PR #323, opened 2021-09-07
Typed Event: Add capability checks to the TCK

timothyjward:
Improving TCK coverage as per issue #322

### DIFF
--- a/org.osgi.test.cases.typedevent/bnd.bnd
+++ b/org.osgi.test.cases.typedevent/bnd.bnd
@@ -9,9 +9,9 @@ Import-Package: ${-signaturetest}, *
 -buildpath: \
 	org.osgi.test.support;version=project,\
 	org.osgi.service.typedevent;version=latest,\
-	org.osgi.resource;version=latest,\
-	org.osgi.namespace.implementation;version=latest,\
-	org.osgi.namespace.service;version=latest,\
+	org.osgi.resource;maven-scope=provided;version=1.0,\
+	org.osgi.namespace.implementation;maven-scope=provided;version=1.0,\
+	org.osgi.namespace.service;maven-scope=provided;version=1.0,\
 	org.osgi.framework;maven-scope=provided;version=1.8,\
 	org.osgi.dto;maven-scope=provided;version=1.0
 
--- a/org.osgi.test.cases.typedevent/bnd.bnd
+++ b/org.osgi.test.cases.typedevent/bnd.bnd
@@ -9,6 +9,9 @@ Import-Package: ${-signaturetest}, *
 -buildpath: \
 	org.osgi.test.support;version=project,\
 	org.osgi.service.typedevent;version=latest,\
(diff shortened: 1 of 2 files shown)


timothyjward:
The TCK failure for this PR was expected and the implementation is now updated.

org.osgi.test.cases.typedevent/bnd.bnd
```diff
@@ -9,6 +9,9 @@ Import-Package: ${-signaturetest}, *
 -buildpath: \
 	org.osgi.test.support;version=project,\
 	org.osgi.service.typedevent;version=latest,\
+	org.osgi.resource;version=latest,\
```
bjhargrave, line 12: Please use actual version numbers for dependent APIs. Preferable the oldest version that satisfies your needs. Only the API under test should use version=latest.

The point is that the test jar should not impose versions higher than what is needed.

Also, these should all be `maven-scope=provided` since they are not needed at runtime or are provided by the OSGi framework impl.

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventCapabilityTestCase.java
```diff
@@ -0,0 +1,111 @@
...
+import org.osgi.framework.wiring.BundleWiring;
+import org.osgi.service.typedevent.TypedEventBus;
+import org.osgi.service.typedevent.monitor.TypedEventMonitor;
+import org.osgi.test.common.annotation.InjectBundleContext;
+import org.osgi.test.junit5.context.BundleContextExtension;
+import org.osgi.test.junit5.service.ServiceExtension;
+
+@ExtendWith(BundleContextExtension.class)
+@ExtendWith(ServiceExtension.class)
+@Nested
```
bjhargrave, line 46: This test class is not nested. So I don't think this annotation is correct.

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventCapabilityTestCase.java
```diff
@@ -0,0 +1,111 @@
...
+
+		assertThat(cap).isNotNull();
+
+		assertThat(cap.getDirectives().get(CAPABILITY_USES_DIRECTIVE))
+				.isEqualTo(clz.getPackage().getName());
+	}
+
+	@Test
+	public void test_bus_service_capability() {
+		testServiceCapability(TypedEventBus.class);
```
bjhargrave, line 79: You should instead consider using parameterized tests.

https://github.com/osgi/osgi/blob/b618b4b9162d057acf46aceb903a52642d57c245/org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventBusTest.java#L108-L118

bjhargrave (changes requested):

PR #324, opened 2021-09-07
[Typed Event] additional Tests

stbischof:


cnf/ext/central.mvn
```diff
@@ -192,3 +192,8 @@ org.slf4j:log4j-over-slf4j:1.7.7
 org.apache.zookeeper:zookeeper:3.4.13
 org.apache.aries.component-dsl:org.apache.aries.component-dsl.component-dsl:1.2.2
 org.apache.aries.typedevent:org.apache.aries.typedevent.bus:0.0.1-SNAPSHOT
+
+org.mockito:mockito-core:jar:3.10.0
```
bjhargrave, line 196: If we are going to add mockito to the build, then we need to add it in a way so that all TCK projects can benefit. 

See https://github.com/osgi/osgi/pull/312 where I added awaitility as a TCK support jar and model adding mockito this way.

org.osgi.test.cases.typedevent/bnd.bnd
```diff
@@ -9,8 +9,9 @@ Import-Package: ${-signaturetest}, *
 -buildpath: \
 	org.osgi.test.support;version=project,\
 	org.osgi.service.typedevent;version=latest,\
-	org.osgi.framework;maven-scope=provided;version=1.8,\
-	org.osgi.dto;maven-scope=provided;version=1.0
+	org.osgi.framework;maven-scope=provided;version='1.8',\
+	org.osgi.dto;maven-scope=provided;version='1.0',\
+	org.mockito.mockito-core
```
bjhargrave, line 14: You don't need to add mockito here since this project already includes osgi.tck.junit-platform.bnd

org.osgi.test.cases.typedevent/bnd.bnd
```diff
@@ -20,3 +21,7 @@ Import-Package: ${-signaturetest}, *
 -runproperties = ${runproperties}
 
 -signaturetest                      = org.osgi.service.typedevent
+
+#for mockito
+-runproperties: \
```
bjhargrave, line 26: This should be done in osgi.tck.junit-platform.bnd using `-runproperties.osgi.tck`.

bjhargrave (changes requested):

cnf/ext/central.mvn
stbischof: I extracted versions into junit.bnd

org.osgi.test.cases.typedevent/bnd.bnd
stbischof: thank you!

org.osgi.test.cases.typedevent/bnd.bnd
stbischof: moved this. thank you

bjhargrave (commented):
LGTM

PR #334, opened 2021-09-14
Update to the latest Aries RSA SNAPSHOT

timothyjward:


timothyjward:
I've fixed a few regressions in Aries RSA, which made 1.16.1-SNAPSHOT pass the TCK, but still not reliably. The issues seemed to be related to the liveness and timing of discovery, so I therefore implemented a discovery provider based on mDNS with JAX-RS/SSE event publication. This is also available as a SNAPSHOT, and seems to make the RSA tests pass.

The new Discovery provider has no unit tests (they're pretty hard to test at the best of times) so it may be that more work is needed in Aries before a release can happen. I've hit the end of the time I can spend on this for now, but hopefully it has moved us closer to our goal.

cnf/repo/org.osgi.impl.service.remoteserviceadmin/org.osgi.impl.service.remoteserviceadmin-7.0.0.lib
```diff
@@ -1,15 +1,17 @@
 org.osgi.service.remoteserviceadmin; version=latest
 org.osgi.service.cm; version=latest
 org.osgi.impl.service.configurator;version=latest
+org.osgi.impl.service.jaxrs;version=latest
```
bjhargrave, line 4: Thie `org.osgi.impl.service.remoteserviceadmin-7.0.0.lib` file should be renamed to end in `-8.0.0.lib`.

bjhargrave (commented):
Thanks for the work!

We will need to move to at least an RC1 type version of the jars in Maven Central as we cannot use SNAPSHOT versions for a compatible implementation.

PR #340, opened 2021-09-16
CM-TCK use osgi-test support

stbischof:
I am using Ubuntu. TCK did not run  befor usage of `osgi-tests` now it runs locally without issues

- Testing wheter this runs on  ci
- what would be the best way to add `osgi-test.junit5.cm` bundle  and blowing all other test away.

### DIFF
--- a/cnf/repo/org.osgi.impl.service.cm/org.osgi.impl.service.cm-7.0.0.lib
+++ b/cnf/repo/org.osgi.impl.service.cm/org.osgi.impl.service.cm-7.0.0.lib
@@ -1,1 +1,2 @@
 org.apache.felix.configadmin; version=1.9
+org.osgi.test.junit5.cm; version=latest
--- a/org.osgi.test.cases.cm/bnd.bnd
+++ b/org.osgi.test.cases.cm/bnd.bnd
@@ -30,19 +30,23 @@ Import-Package: ${-signaturetest}, *
 	
 -signaturetest                      = org.osgi.service.cm
 
--buildpath = \
-    org.osgi.test.support;version=project, \
-    org.osgi.framework;maven-scope=provided;version=1.8,\
-    org.osgi.resource;maven-scope=provided;version=1.0,\
-    org.osgi.service.permissionadmin;maven-scope=provided;version=1.2,\
-    org.osgi.namespace.implementation;version=1.0,\
-    org.osgi.namespace.service;version=1.0,\
-    org.osgi.service.coordinator;version=1.0,\
-	org.osgi.service.cm;version=latest
+-buildpath: \
+	org.osgi.test.support;version=project,\
+	org.osgi.framework;maven-scope=provided;version='1.8',\
+	org.osgi.resource;maven-scope=provided;version='1.0',\
+	org.osgi.service.permissionadmin;maven-scope=provided;version='1.2',\
+	org.osgi.namespace.implementation;version='1.0',\
+	org.osgi.namespace.service;version='1.0',\
+	org.osgi.service.coordinator;version='1.0',\
+	org.osgi.service.cm;version=latest,\
+	org.osgi.annotation.versioning,\
+	org.osgi.test.junit5.cm
 
 -runbundles = \
 	org.osgi.impl.service.log; version=latest, \
 	org.osgi.impl.service.cm; version=latest, \
-	org.osgi.impl.service.coordinator; version=latest
+	org.osgi.impl.service.coordinator; version=latest,\
+	org.osgi.test.junit5.cm; version=latest
+	
 	
 -runproperties = ${runsecureproperties}
--- a/org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMControl.java
+++ b/org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMControl.java
@@ -17,9 +17,23 @@
  *******************************************************************************/
 package org.osgi.test.cases.cm.junit;
 
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 import static org.osgi.test.cases.cm.shared.Util.singletonDictionary;
+import static org.osgi.test.support.compatibility.DefaultTestBundleControl.assertException;
+import static org.osgi.test.support.compatibility.DefaultTestBundleControl.failException;
+import static org.osgi.test.support.compatibility.DefaultTestBundleControl.pass;
+import static org.osgi.test.support.compatibility.DefaultTestBundleControl.trace;
+import static org.osgi.test.support.sleep.Sleep.sleep;
 
 import java.io.IOException;
+import java.lang.reflect.Method;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -32,6 +46,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.PropertyPermission;
 import java.util.Set;
 import java.util.concurrent.BrokenBarrierException;
@@ -39,8 +54,14 @@ import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.osgi.framework.AdminPermission;
 import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
 import org.osgi.framework.BundleException;
 import org.osgi.framework.Constants;
 import org.osgi.framework.InvalidSyntaxException;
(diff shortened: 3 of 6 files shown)


org.osgi.test.cases.framework/src/org/osgi/test/cases/framework/junit/context/InstallBundleTest.java
```diff
@@ -0,0 +1,64 @@
+/*******************************************************************************
```
tjwatson, line 1: What does this have to do with the CM TCK?

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/ConfigurationPluginTests.java
```diff
@@ -87,21 +115,25 @@ protected void setUp() throws Exception {
         }
 	}
 
+	private long getLongProperty(String string, int i) {
```
tjwatson, line 118: What is this used for?

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
```diff
@@ -345,4 +372,8 @@ public void updated(Dictionary<String, ? > properties)
 		assertFalse(events.get(3));
 
 	}
+
```
tjwatson, line 375: Why do we have this?

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/ConfigurationPluginTests.java
```diff
@@ -250,14 +292,15 @@ public void testRankingForManagedService() throws Exception {
...
-		cm.getConfiguration(pid).delete();
+		cm.waitForService(500).getConfiguration(pid).delete();
 		trace("Wait until the ManagedService has gotten the delete");
 
 		assertTrue("Update done",
 				synchronizer.waitForSignal(SIGNAL_WAITING_TIME));
 		assertNull(synchronizer.getProps());
 		assertTrue(context.invocationOrder.isEmpty());
 		assertTrue(context.errors.isEmpty());
+		reg.unregister();
```
tjwatson, line 303: This service remaining in the registry may be the root cause of the issues

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/ConfigurationPluginTests.java
```diff
@@ -335,15 +380,19 @@ public void testRankingForManagedServiceFactory() throws Exception {
...
```
tjwatson, line 383: This service in the registry may be the root cause of the issues.

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/ConfigurationPluginTests.java
```diff
@@ -362,15 +412,18 @@ public void testTargettingManagedService() throws Exception {
 		assertEquals("val", props.get("key"));
 		verifyPlugins(props, false, Arrays.asList(1, 2), Arrays.asList(1, 2),
 				context);
+		reg.unregister();
```
tjwatson, line 415: Same comment, good find!

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/ConfigurationPluginTests.java
```diff
@@ -390,15 +444,17 @@ public void testTargettingManagedServiceFactory() throws Exception {
 		assertEquals("val", props.get("key"));
 		verifyPlugins(props, true, Arrays.asList(1, 2), Arrays.asList(1, 2),
 				context);
+		reg.unregister();
```
tjwatson, line 447: +1

tjwatson (commented):
Overall this looks like a really good direction.  I think we should consider this for an R8 fix.

org.osgi.test.cases.framework/src/org/osgi/test/cases/framework/junit/context/InstallBundleTest.java
stbischof: Nothing.

I just had some errors and had to find out whether this ins an tck Design error or something that came from Framework...

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/ConfigurationPluginTests.java
stbischof: The original test hat optional to set the sleep time.. this is just something i will clean up when tests work every time

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/ConfigurationPluginTests.java
stbischof: Yes.  The unregisting of the managed services is an issue. I expected that the closeable bundlecontext of osgi-tests will unregister this. Bur this did not happed. Need more investigation here

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
stbischof: sorry i reset the commit. so i do not know what happened here

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMControl.java
```diff
@@ -2294,10 +2362,10 @@ public void updated(final Dictionary<String, ?> properties) throws Configuration
 			assertNull(updates.get(0));
 			assertEquals("somevalue", updates.get(1).get("scalar"));
 			assertEquals("newvalue", updates.get(2).get("scalar"));
-		} finally {
-			this.unregisterService(listener);
-			this.unregisterService(ms);
-		}
+			reg.unregister();
```
stbischof, line 2365: same here @tjwatson

cnf/repo/org.osgi.impl.service.cm/org.osgi.impl.service.cm-7.0.0.lib
```diff
@@ -1 +1,2 @@
 org.apache.felix.configadmin; version=1.9
+org.osgi.test.junit5.cm; version=latest
```
bjhargrave, line 2: A testing jar must not be in the impl lib file!

org.osgi.test.cases.cm/bnd.bnd
```diff
@@ -30,19 +30,23 @@ Import-Package: ${-signaturetest}, *
...
+-buildpath: \
+	org.osgi.test.support;version=project,\
+	org.osgi.framework;maven-scope=provided;version='1.8',\
+	org.osgi.resource;maven-scope=provided;version='1.0',\
+	org.osgi.service.permissionadmin;maven-scope=provided;version='1.2',\
+	org.osgi.namespace.implementation;version='1.0',\
+	org.osgi.namespace.service;version='1.0',\
+	org.osgi.service.coordinator;version='1.0',\
+	org.osgi.service.cm;version=latest,\
+	org.osgi.annotation.versioning,\
```
bjhargrave, line 42: Why does the TCK need versioning annotations? And it does need them, it should use `osgi.annotations;version=8.0`.

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMControl.java
```diff
@@ -217,29 +254,43 @@
 
 	private static final String neverlandLocation = "http://neverneverland/";
 
-	protected void setUp() throws Exception {
+	@BeforeEach
+	void setUp(TestInfo testInfo) throws Exception {
+		name = testInfo.getTestMethod()
+				.map(Method::getName)
+				.orElse("AutoName_" + System.currentTimeMillis());
```
bjhargrave, line 261: use `orElseGet` to avoid the string building when unnecessary.

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMControl.java
```diff
@@ -311,17 +361,19 @@ private void printoutPermissions() {
 	}
 
 	private void setBundlePermission(Bundle b, List<PermissionInfo> list) {
-		if (permAdmin == null)
+		if (permAdminA.getService() == null)
```
bjhargrave, line 364: You should put `permAdminA.getService()` into a variable and then test for `null` and use. Calling `permAdminA.getService()` twice could give different results.

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMControl.java
```diff
@@ -2294,10 +2367,10 @@ public void updated(final Dictionary<String, ?> properties) throws Configuration
 			assertNull(updates.get(0));
 			assertEquals("somevalue", updates.get(1).get("scalar"));
 			assertEquals("newvalue", updates.get(2).get("scalar"));
-		} finally {
-			this.unregisterService(listener);
-			this.unregisterService(ms);
-		}
+			reg.unregister();
+			// TODO BJ:WHY MUST THIS BE DONE MANUALLY????
```
bjhargrave, line 2371: I assume that since the test method uses the raw service API to register a service, it must then unregister it as no one else will.

bjhargrave (commented):
Wow! There is a lot of nice work in here. @tjwatson Do you think this is worth merging at this point in the release?

cnf/repo/org.osgi.impl.service.cm/org.osgi.impl.service.cm-7.0.0.lib
stbischof: sure but if i put it into the junit.tck all other test will fail because of the  missing cm and adding it just to the bndrun file just works in ide but not on ci because it does not find it on one of the given repos.

bjhargrave: There is no reason some other project using the CM impl needs this osgi-test jar. You already have it in the -runbundles and -buildpath of the TCK which should be sufficient. Yes, you cannot put it in the osgi.tck.junit-platform.bnd file.

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
tjwatson: I was asking about the getLongProperty method you had added as a stub method, but it is gone now.

tjwatson:
> Wow! There is a lot of nice work in here. @tjwatson Do you think this is worth merging at this point in the release?

Only if we have some proof that it makes the TCK pass more reliably.  But so far I have not seen that.  I really think we may be chasing a bug in Felix CM implementation at this point.

PR #343, opened 2021-09-17
CM TCK - configure coordinated ManagedService

stbischof:
Additional tests on cm in combination with the coordinator. Currently the RI fails on  `test_deliver_existing_Configuration_to_later_registered_ManagedService`.

LOCAL BUILD:
At the moment the update method of the Managed Service is  called 4 Times.
1. Dictionary with initial Config
2. null  **<--i think this is wrong**
3. Dictionary with the updated config
4. null , because removed

CI BUILD:
non of the 3 events happened `: expected:<3> but was:<0>`

From the point of view of the managed service, the service must `thing` that it is
1. configured
2.  removed
3. configured again
4. removed again

The test `testCoordinatedConfigurationOnBeforeRegisteredManagedService` runs on my local machine without an error.
https://issues.apache.org/jira/browse/FELIX-6470

### DIFF
--- a/org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
+++ b/org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
@@ -270,6 +270,139 @@ public class CMCoordinationTestCase extends DefaultTestBundleControl {
 
 	}
 
+	// Test opposite of:
+	// If a Managed Service is registered and no configuration information is
+	// available or the configuration is not visible then the Configuration
+	// Admin service must call updated(Dictionary) with a null parameter.
+	//
+	// Configuration exists before ManagedService is registered. No update(null)
+	// use cfg-dictionary directly
+
+	public void test_deliver_existing_Configuration_to_later_registered_ManagedService()
+			throws Exception {
+		// start a coordination
+		final Coordinator c = this.getService(Coordinator.class);
+		final Coordination coord = c.begin("cm-test1", 0);
+		final List<Boolean> events = new ArrayList<>();
+
+		final String pid = this.getClass().getName() + ".mstestpid2";
+
+		try {
+			// create the configuration
+			final Dictionary<String,Object> props = new Hashtable<>();
+			props.put("key", "value");
+
+			final Configuration conf = this.cm.getConfiguration(pid);
+			conf.update(props);
+
+			// add managed service
+			final Dictionary<String,Object> msProps = new Hashtable<>();
+			msProps.put(Constants.SERVICE_PID, pid);
+
+			this.registerService(ManagedService.class.getName(),
+					new ManagedService() {
+
+						@Override
+						public void updated(Dictionary<String, ? > properties)
+								throws ConfigurationException {
+							System.out.println(properties);
+							events.add(properties != null);
+						}
+
+					}, msProps);
+
+			sleep();
+			assertEquals(0, events.size());
+
+			// update configuration
+			props.put("key2", "value2");
+			conf.update(props);
+
+			sleep();
+			assertEquals(0, events.size());
+
+			// delete configuration
+			conf.delete();
+
+			sleep();
+			assertEquals(0, events.size());
+		} finally {
+			coord.end();
+		}
+
+		// wait and verify listener
+		sleep();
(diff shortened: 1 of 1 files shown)


tjwatson:
Here the coordination exists on the thread where you are registering the service.  At that time configuration admin treats it as if there is no configuration yet.  Perhaps at this point it is doing an async delivery of `null` to the just registered ManagedService (although I would think that happens synchronously, but I'm not sure if the spec mandates that).  Then the thread continues and does an update and removal while the coordination is still active.  That queues up 3 events that should happen when the coordination ends.  Finally the coordination ends causing the 3 events to be fired to the ManagedService.  Perhaps these events are being queued in such a way that they are getting interleaved with the original `null` event from service registration?

stbischof:
The `org.apache.felix.cm.impl.helper.ManagedServiceTracker` tracks the added Service and gives the Configuration to the ConfigurationManager. This one knows about the coordination and queues the configuration. Without checking the existence of an other configuration that is queued before. 

Stack:
```
Thread [main] (Suspended (breakpoint at line 76 in CoordinatorUtil))	
	CoordinatorUtil.addToCoordination(Object, UpdateThread, Runnable) line: 76	
	ConfigurationManager.configure(String[], ServiceReference, boolean, ConfigurationMap<?>) line: 785	
	ManagedServiceTracker(BaseTracker<S>).configure(ServiceReference<S>, String[], ConfigurationMap<?>) line: 111	
	ManagedServiceTracker(BaseTracker<S>).addingService(ServiceReference<S>) line: 78	
	ManagedServiceTracker(BaseTracker<S>).addingService(ServiceReference) line: 54	
	ServiceTracker$Tracked.customizerAdding(ServiceReference<S>, ServiceEvent) line: 943	
	ServiceTracker$Tracked.customizerAdding(Object, Object) line: 1	
	ServiceTracker$Tracked(AbstractTracked<S,T,R>).trackAdding(S, R) line: 256	
	ServiceTracker$Tracked(AbstractTracked<S,T,R>).track(S, R) line: 229	
	ServiceTracker$Tracked.serviceChanged(ServiceEvent) line: 903	
	FilteredServiceListener.serviceChanged(ServiceEvent) line: 120	
	BundleContextImpl.dispatchEvent(Object, Object, int, Object) line: 957	
	EventManager.dispatchEvent(Set<Entry<K,V>>, EventDispatcher<K,V,E>, int, E) line: 234	
	ListenerQueue<K,V,E>.dispatchEventSynchronous(int, E) line: 151	
	ServiceRegistry.publishServiceEventPrivileged(ServiceEvent) line: 936	
	ServiceRegistry$1.run() line: 878	
	ServiceRegistry$1.run() line: 1	
	AccessController.executePrivileged(PrivilegedAction<T>, AccessControlContext, Class<?>) line: 753	
	AccessController.doPrivileged(PrivilegedAction<T>) line: 312	
	ServiceRegistry.publishServiceEvent(ServiceEvent) line: 875	
	ServiceRegistrationImpl<S>.register(Dictionary<String,?>) line: 141	
	ServiceRegistry.registerService(BundleContextImpl, String[], Object, Dictionary<String,?>) line: 261	
	BundleContextImpl.registerService(String[], Object, Dictionary<String,?>) line: 496	
	BundleContextImpl.registerService(String, Object, Dictionary<String,?>) line: 515	
	CMCoordinationTestCase(DefaultTestBundleControl).registerService(String, Object, Dictionary<String,?>) line: 340	



```

bjhargrave:
I don't think we want to put this issue in the way of completing Compendium R8.



> Perhaps at this point it is doing an async delivery of `null` to the just registered ManagedService (although I would think that happens synchronously, but I'm not sure if the spec mandates that).

According to the CM spec:

> If configurations are created, updated or deleted and an implicit coordination exists, the Configuration Admin service must delay notifications until the coordination terminates. However the configuration changes must be persisted immediately.

Since the Configuration actually exists at the time the ManagedServices is registered (even though a Coordination is in effect), I don't see that CM should send a null other than the final null when the Configuration is deleted.

stbischof:
Related Issue in Felix-Jira: https://issues.apache.org/jira/browse/FELIX-6470

stbischof:
This is a valid tck-test. So we could merge this.
It is on felix-cm site to fix this.

bjhargrave:
> So we could merge this.

Can you rebase on main?

PR #363, opened 2021-09-29
ci: Add step for release to create a staging repository id

bjhargrave:
We use a curl request to obtain the staging profile URI for the
org.osgi group id. And then we make a curl request to start a
staging repository. The URL to the created staging repository
is place in the GitHub actions env context for access by later
steps in the job.

### DIFF
--- a/.github/scripts/staging.sh
+++ b/.github/scripts/staging.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+#*******************************************************************************
+# Copyright (c) Contributors to the Eclipse Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


juergen-albert:
Looking at the script, I'm not sure if all of this is really necessary. I may have missed some requirements here that made it necessary to manually start a staging repository. When we deploy our gecko bundles to maven central we never have to jump through all these hoops, because sonartype starts a staging repo automatically.

In our case we only include the following bnd file if we want to do a release:

```
-plugin.sonatype: \
	aQute.bnd.repository.maven.provider.MavenBndRepository;\
		snapshotUrl=https://oss.sonatype.org/content/repositories/snapshots/;\
		releaseUrl=https://oss.sonatype.org/service/local/staging/deploy/maven2/;\
		index=${.}/sonatype.maven;\
		name='Sonatype'
				
-releaserepo.sonatype: Sonatype
```

bjhargrave:
That can work fine but sometimes OSSRH splits the deployed artifacts across multiple staging repositories as happened with the RC2 build I made a few days ago as a test. Those who were able to attend the project call yesterday saw this and we discussed it. @rotty3000 found the information to add these commands to start a staging repository so we can use that staging repositories deployment url to avoid multiple staging repositories. It worked correctly in my test last night. Thanks @rotty3000.

So yes, it really is necessary.

PR #393, opened 2021-12-23
Make DTOUtil stop relying on exceptions

### LINKED ISSUE #392: Converter's DTOUtil is overly reliant on exceptions for normal execution flow
DTOUtil.isDTOType() [1] relies on exceptions to find out whether a class has a public constructor and to find out if it has any methods - as a DTO should have a public no-args constructor and no methods beyond the ones defined by Object.class. 

This works but is wasteful and slow. We should implement a different strategy to identify a DTO class as such.

[1] https://github.com/osgi/osgi/blob/main/org.osgi.util.converter/src/org/osgi/util/converter/DTOUtil.java

bosschaert:
Fixes #392

### DIFF
--- a/org.osgi.util.converter/src/org/osgi/util/converter/DTOUtil.java
+++ b/org.osgi.util.converter/src/org/osgi/util/converter/DTOUtil.java
@@ -37,30 +37,20 @@ class DTOUtil {
 
 	static boolean isDTOType(Class< ? > cls, boolean ignorePublicNoArgsCtor) {
 		if (!ignorePublicNoArgsCtor) {
-			boolean hasPublicNoargsCtor = false;
-			for (Constructor< ? > ctor : cls.getConstructors()) {
-				if (ctor.getParameterCount() == 0) {
-					hasPublicNoargsCtor = true;
-					break;
-				}
-			}
-			if (!hasPublicNoargsCtor) {
+			if (Arrays.stream(cls.getConstructors())
+					.noneMatch(ctor -> ctor.getParameterCount() == 0)) {
+				// No public zero-arg constructor, not a DTO
 				return false;
 			}
 		}
 
 		for (Method m : cls.getMethods()) {
-			boolean objectClassMethod = false;
-			for (Method om : OBJECT_CLASS_METHODS) {
-				if (om.getName().equals(m.getName())) {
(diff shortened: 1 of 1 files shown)


karlpauls (approved):
LGTM - thanks!

bosschaert:
I force-pushed the commit now with the signoff and my registered email address.

org.osgi.util.converter/src/org/osgi/util/converter/DTOUtil.java
```diff
@@ -18,45 +18,53 @@
...
-			} catch (NoSuchMethodException | SecurityException e) {
-				// No public zero-arg constructor, not a DTO
+			boolean hasPublicNoargsCtor = false;
+			for (Constructor< ? > ctor : cls.getConstructors()) {
+				if (ctor.getParameterCount() == 0) {
+					hasPublicNoargsCtor = true;
+					break;
+				}
+			}
+			if (!hasPublicNoargsCtor) {
```
bjhargrave, line 47:
```suggestion
			if (Arrays.stream(cls.getConstructors())
					.noneMatch(ctor -> ctor.getParameterCount() == 0)) {
				// No public zero-arg constructor, not a DTO
```
More concise alternative.

org.osgi.util.converter/src/org/osgi/util/converter/DTOUtil.java
```diff
@@ -18,45 +18,53 @@
...
+			for (Method om : OBJECT_CLASS_METHODS) {
+				if (om.getName().equals(m.getName())) {
+					if (Arrays.equals(om.getParameterTypes(),
+							m.getParameterTypes())) {
+						objectClassMethod = true;
+						break;
+					}
+				}
+			}
+			if (!objectClassMethod) {
```
bjhargrave, line 63:
```suggestion
			if (Arrays.stream(OBJECT_CLASS_METHODS)
					.noneMatch(om -> om.getName().equals(m.getName())
							&& Arrays.equals(om.getParameterTypes(),
									m.getParameterTypes()))) {
				// Not a method defined by Object.class (or override of such
				// method)
```

More concise alternative.

bosschaert:
Sure we can use a stream based variant too. Thanks @bjhargrave !

bjhargrave (approved):

PR #399, opened 2022-01-17
Add requirement about feature runtime introspection

bosschaert:
Signed-off-by: David Bosschaert <david.bosschaert@gmail.com>


.design/requirements-386.md
```diff
@@ -53,3 +53,4 @@ Bill has a Feature that contains a custom extension to initialize the database f
 * FL-0070: The solution MUST provide a predictable mechanism to merge multiple Features, taking into account that bundles, configurations and custom extensions each have different requirements on how they are merged.
 * FL-0080: The solution MUST provide a mechanism to resolve merge conflicts.
 * FL-0090: The solution MUST provide a well defined mechanism for handling failures when merging features or launching them.
+* FL-0100: The solution SHOULD provide a runtime service where the currently launched features can be inspected.
```
bjhargrave, line 56: SHOULD seems too weak here. If this is not MUST, then no one can rely upon the runtime servivce.

Perhaps:

FL-0100: The solution MUST provide a well defined mechanism where the currently launched features can be inspected.

bosschaert: Works for me 😄  I've updated the PR

PR #418, opened 2022-04-04
[tck] org.osgi.service.jdbc - enhancements

stbischof:
* switch to junit 5
* use osgi-test
* remove tests on custom properties ("junk" -- derby related test)
* test (notNull) return value of DataSource and Driver methods

Discussion:
Spec uses `should` and `optional` this would be a way to test this:
* test "should" with separate @Tag and "@EnabledIfSystemProperty"

### DIFF
--- a/org.osgi.test.cases.jdbc/bnd.bnd
+++ b/org.osgi.test.cases.jdbc/bnd.bnd
@@ -18,3 +18,5 @@ Import-Package: ${-signaturetest}, *
 	org.osgi.impl.bundle.derby; version=latest
 
 -signaturetest                      = org.osgi.service.jdbc
+
+-runproperties.tck.cardinality: org.osgi.tck.cardinality=SHOULD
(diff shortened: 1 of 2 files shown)


timothyjward:
> Discussion:
Spec uses should and optional this would be a way to test this:

> test "should" with separate @tag and "@EnabledIfSystemProperty"

This actually raises a long-standing issue with the service, which is that there is no way for the client to know if a method is implemented or not. Would it make more sense to define some service properties on the DataSourceFactory service so that everyone can know what's supposed to work?

stbischof:
It would ! It just getting to late yesterday. And some thing i would not address in this PR because this should just fix some issues I had on the implementation site.

There are two topics:
- SHOULD and OPTIONAL sentences in the Spec -> @Tag ed Testmethods and "@EnabledIfSystemProperty"
- Unclear which method can be used without exception.

Service-Properties should be specified in an 1.1 version.
e.g.

```
org.osgi.service.jdbc.capability={ "DataSource", "ConnectionPooledDataSource", "XADataSource", "Driver"}
```
or
```
org.osgi.service.jdbc.capability.dataSource=true
org.osgi.service.jdbc.capability.ConnectionPooledDataSource=true
org.osgi.service.jdbc.capability.XADataSource=true
org.osgi.service.jdbc.capability.Driver=true
```

PR #423, opened 2022-04-19
[jaxrs] Minor TCK fixes to improve implementation compatibility

timothyjward:
The JAX-RS TCK makes heavy use of String concatenation to build URIs, but it isn't very careful about introducing extra / characters. Extra / characters can cause failures in some JAX-RS clients so we should use the built in API methods to avoid these problema

Signed-off-by: Tim Ward <timothyjward@apache.org>

### DIFF
--- a/org.osgi.test.cases.jaxrs/src/org/osgi/test/cases/jaxrs/junit/ClientTestCase.java
+++ b/org.osgi.test.cases.jaxrs/src/org/osgi/test/cases/jaxrs/junit/ClientTestCase.java
@@ -64,7 +64,7 @@ public class ClientTestCase extends AbstractJAXRSTestCase {
 
 		String baseURI = getBaseURI();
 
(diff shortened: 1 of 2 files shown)


maho7791 (commented):
Is the leading / in the SSETestCase path in line 62 correct? I believe we have to remove it here as well

timothyjward:
> Is the leading / in the SSETestCase path in line 62 correct? I believe we have to remove it here as well

I'll remove it, but the definition of the [`path`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/WebTarget.html#path-java.lang.String-) method requires it to handle leading/trailing `/` characters properly. Previously we were concatenating `String` objects which was the source of the problem.

maho7791:
looks good

bjhargrave (approved):

Issue #426, opened 2022-04-27
TCK JaxRS Whiteboard SSE Test is not JaxRs Spec compliant


maho7791:
Test:
https://github.com/osgi/osgi/blob/main/org.osgi.test.cases.jaxrs/src/org/osgi/test/cases/jaxrs/junit/SSETestCase.java

The test setup expects that the client stops receiving data, when the server side closes the event sink. According to the JaxRs specification, the server side close is handles as a connection loss. The client has to re-establish the connection. So the setup will:
* receive more than the 10 data elements, because a reconnect happens
* the onComplete that releases the sempahore is not called
I stumbled up that while testing our JaxRs Whiteboard against the TCK with Jersey. 

There is already an issue at Jersey, that describes the problem:
https://github.com/eclipse-ee4j/jersey/issues/4996

maho7791:
See PR https://github.com/osgi/osgi/pull/427

timothyjward:
Fixed by #428

PR #428, opened 2022-04-27
[jaxrs] Improve the TCK compatibility with the JAX-RS specification

timothyjward:
The JAX-RS SseEventSource is somewhat poorly specified, in that it will continually reconnect at the end of a stream, and is not required to notify of completion in this case. The server is expected to send a 204 response if the client is to stop connecting, at which point the client is notified that the stream has ended.

This change also adds an explicit message to set the client reconnect time to be one 'tick' which is 500 ms, ensuring that default delays in the client will not cause the test to time out.

Signed-off-by: Tim Ward <timothyjward@apache.org>

### DIFF
--- a/org.osgi.test.cases.jaxrs/src/org/osgi/test/cases/jaxrs/junit/SSETestCase.java
+++ b/org.osgi.test.cases.jaxrs/src/org/osgi/test/cases/jaxrs/junit/SSETestCase.java
@@ -95,7 +95,7 @@ public class SSETestCase extends AbstractJAXRSTestCase {
 
 			assertNull(ref.get());
 
-			assertEquals(Arrays.asList(0, 1, 2, 3, 4, 5, 6, 7, 8, 9), list);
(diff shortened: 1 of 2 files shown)


timothyjward:
This is an alternative fix to #426 which updates the resource method to use JAX-RS specified behaviour to ensure a clean shutdown of the client, rather than closing it in the test. It also protects against situations where the stream is interrupted part way through.

There are three parts to this:

1. Check the value of the last event id header. If it is greater than or equal to the last event (9) then return a 204 to close the client
2. If there is no last event id then set a reconnect time in a hello event, this guarantees that the client will reconnect after 500ms
3.  Stream events starting from one after the last received event, or the first event if no last event defined

maho7791:
Do we need the 'hello' stuff? The logic is also working without.

timothyjward:
> Do we need the 'hello' stuff? The logic is also working without.

We need a message to hang the retry timeout setting on, otherwise the client might set a long value. The other option is for the test code to set it when creating the event source.

PR #429, opened 2022-04-27
[tck jdbc] do expect an exception on not defines properties

stbischof:
The TCK expects that the implementation throws an Exception using the property `junk`. we cant expect this from all implementations. They also could ignore undefined properties.

Spec:
Which property keys and values are supported depends on the driver
implementation. Drivers can support additional custom configuration
properties.

Signed-off-by: Stefan Bischof <stbischof@bipolis.org>

### DIFF
--- a/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
+++ b/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
@@ -91,12 +91,6 @@ public class JDBCTestCase extends OSGiTestCase {
 			return;
 		}
 
-		// make sure we get an exception if we use an unknown property
-		props.put( "junk", "junk" );
-		try {
-			factory.createDataSource( props );
-			fail("Should have gotten a SQLException for invalid property \"junk\".");
-		} catch ( SQLException ignore ) { }
 	}
 
 	public void testCreateConnectionPoolDataSource() throws Exception {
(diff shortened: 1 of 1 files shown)


bjhargrave (approved):

laeubi:
Just wondering, would this not also require to adjust the "spec text" some where? 

If I look at https://osgi.github.io/osgi/cmpn/service.jdbc.html it says:

> If the property cannot be set on the DataSource/XADataSource/ConnectionPoolDataSource/Driver/ being created then a SQLException must be thrown.

stbischof:
The test this PR removed expected that no implementation will ever accept the property junk. 

In my opinion there is a difference between 
- cannot be set
- cannot be processes

If an Implementation accepts also properties and does not throw any exception but also does not use the Property then it means to me, that the Property is set.

laeubi:
> In my opinion there is a difference between
>     * cannot be set
>     * cannot be processes

Sure but the spec is not clear and the TCK seem to assume that these are equal things. One might argue that "cannot be set" is something like an integer is required but a string is given, I also won't say just because I have no clue what to do with a property `junk` is the same as "cannot be processes", because ignoring it is also some kind of processing :-)

> For me a MUST be Logged would fit better

I think neither logging nor throwing should be a MUST, it just puts additional burden on the implementation without any gain, because no one would expect that a user uses trial&error to find out supported properties but has to read the docs carefully or even take a look at the implementation if one wants to use any non standard properties, something that 
- https://github.com/osgi/osgi/issues/436

tries to address already, and use something like

> if a property cannot be applied then an SQL exception must be thrown, properties unknown to this DataSource must be ignored.

stbischof:
I removed this because i could not guarantee, that any implementation is able to process `junk`.
It is up to the implementation to decide whether the value is processed or not.

laeubi:
> that any implementation is able to process `junk`.

I think that's rather weak assert there could be any code understanding `junk` :-) if you are really certain one better replace it then with a UUID or something similar random to ensure one do not hits a valid value by accident?

PR #432, opened 2022-04-27
[jdbc] define properties that indicate capabilities of DataSourceFactory

stbischof:
@timothyjward  properties like discussed in #418 ?

### DIFF
--- a/org.osgi.service.jdbc/src/org/osgi/service/jdbc/DataSourceFactory.java
+++ b/org.osgi.service.jdbc/src/org/osgi/service/jdbc/DataSourceFactory.java
@@ -41,6 +41,56 @@ import javax.sql.XADataSource;
  */
 public interface DataSourceFactory {
 	/**
+	 * Service property used by a {@code DataSourceFactory} to declare the
+	 * capabilities when registering a JDBC DataSourceFactory service. Clients
+	 * may filter or test this property to determine if the comparison method
+	 * could be used.
+	 * 
+	 * @since 1.1
+	 */
+	public static final String	OSGI_JDBC_CAPABILITY							= "osgi.jdbc.datasourcefactory.capability";
+
+	/**
+	 * As a value of the property 'osgi.jdbc.datasourcefactory.capability' it
+	 * indicates that the {@code DataSourceFactory} is able to provide a
+	 * {@code Driver} invoking the {@code createDriver(Properties props)}
+	 * method.
+	 * 
+	 * @since 1.1
+	 */
+	public static final String	OSGI_JDBC_CAPABILITY_DRIVER						= "driver";
+
+	/**
+	 * As a value of the property 'osgi.jdbc.datasourcefactory.capability' it
+	 * indicates that the {@code DataSourceFactory} is able to provide a
+	 * {@code DataSource} invoking the
+	 * {@code createDataSource(Properties props)} method.
+	 * 
+	 * @since 1.1
+	 */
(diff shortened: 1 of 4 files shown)


bjhargrave (commented):
Please add `@since 1.1` javadoc tags to each new member.

stbischof:
I also added some Spec-Text a and the implementation part.

test will be done in a separate PR

org.osgi.service.jdbc/src/org/osgi/service/jdbc/DataSourceFactory.java
```diff
@@ -40,6 +40,56 @@
  * @ThreadSafe
  */
 public interface DataSourceFactory {
+	/**
+	 * Service property used by a {@code DataSourceFactory} to declare the
+	 * capabilities when registering a JDBC DataSourceFactory service. Clients
+	 * may filter or test this property to determine if the comparison method
+	 * could be used.
+	 * 
+	 * @since 1.1.0
```
bjhargrave, line 49: Just `1.1`. Not `1.1.0` as we would never change API on a micro version change. So all new API are added on a major or minor change. You will see that we only use major.minor in all other `@since` java tags.

org.osgi.service.jdbc/src/org/osgi/service/jdbc/package-info.java
```diff
@@ -37,7 +37,7 @@
  * @author $Id$
  */
 
-@Version("1.0.1")
+@Version("1.1.0")
```
bjhargrave, line 40: This should be `@Version("1.1")`. We only use the micro component if necessary.

org.osgi.service.jdbc/src/org/osgi/service/jdbc/DataSourceFactory.java
```diff
@@ -40,6 +40,56 @@
...
+
+	/**
+	 * As a value of the property 'osgi.jdbc.datasourcefactory.capability' it
+	 * indicates that the {@code DataSourceFactory} is able to provide a
+	 * {@code XADataSource} invoking the
+	 * {@code createXADataSource(Properties props)} method.
+	 * 
+	 * @since 1.1.0
+	 */
+	public static final String	OSGI_JDBC_CAPABILITY_XADATADOURCE				= "xadatadource";
```
bjhargrave, line 91:
```suggestion
	public static final String	OSGI_JDBC_CAPABILITY_XADATASOURCE				= "xadatasource";
```

osgi.specs/docbook/125/service.jdbc.xml
```diff
@@ -295,10 +295,44 @@ DataSourceFactory dsf = (DataSourceFactory) tracker.getService();</programlistin
...
+          xrefstyle="hyperlink"/></para>
+        </listitem>
+        <listitem>
+          <para><xref
+          linkend="org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_CAPABILITY_CONNECTIONPOOLEDDATASOURCE"
+          xrefstyle="hyperlink"/></para>
+        </listitem>
+        <listitem>
+          <para><xref
+          linkend="org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_CAPABILITY_XADATADOURCE"
```
bjhargrave, line 332:
```suggestion
          linkend="org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_CAPABILITY_XADATASOURCE"
```

bjhargrave (changes requested):

PR #457, opened 2022-07-06
JAX-RS TCK - Fix race condition in the async lifecycle test

timothyjward:
This PR fixes a race condition which causes an intermittent (and incorrect) failure with some JAX-RS whiteboard implementations. I have based it on top of #453, however it could easily be cherry-picked as an independent commit.


bjhargrave (changes requested):
This second commit looks good. I think you need to rebase on main since the first commit in this PR is now already on main.

bjhargrave (approved):

Issue #456, opened 2022-07-04
Explicit timeout handling on Promise API


glimmerveen:
Currently, when a timeout is specified on a promise, the Promise returned by timeout fails with a TimeoutException when the upstream promise did not resolve beforehand. Handling of a timeout relies on the same operations as "regular" failures. 

If in the failure handling you want to distinguish between a timeout or application failure, you'll need to do an `instanceof` check on the _failure_, which hurts code readability a bit. Adding explicit handlers specifically for timeout related failures could improve the readability of application code.

For example, by extending the Promise API with methods like:

* `Promise<T> onTimeout(Runnable callback)`
* `Promise<T> recoverTimeout(Supplier<? extends T> recovery)`
* `Promise<T> recoverTimeoutWith(Supplier<Promise<? extends T>> recovery)`
* `Promise<T> fallbackTimeoutTo(Promise<? extends T> fallback)`

Though I think application code could benefit from this, I do also see some downsides. Primarily on how these proposed methods would coexist with the use of the existing failure handling operations:
* For example, when there is both an `onFailure` and an `onTimeout` specified, when the timeout triggers, both handlers would be called. I guess preventing the onFailure callback in this case would be a backwards incompatible change, and therefor not an option.
* Also the order of operations would matter. Specifying the `recoverWith` before `recoverTimeoutWith`, could render the latter useless when the former would also (inadvertently) handle the TimeoutException case. This is not really new, as the current API also permits multiple `recoverWith` being chained, which also requires care to ensure each function handles the case(s) it is supposed to.

Any thoughts on this? Would such a change be an acceptable extension? Are there perhaps alternative approaches to timeout handling to be considered?

bjhargrave:
I am convinced this is not a good idea. It would require to spec to define two callback queues (for non-timeout callbacks and for timeout callbacks). And then also to call the right call back depending on the exception type of the failure.

The correct answer here is to do the instanceof check you already mention. Timeout is just one of many exception types which can be present in a promise failure. Recovery code may need to take special action for any number of exception types including TimeoutException.

[...]

PR #461, opened 2022-07-14
promise: Overload failure related methods to take a failure type

bjhargrave:
This allows the user to register failure related callback to handle
specific failure types.

Spec updates still needed.

Fixes https://github.com/osgi/osgi/issues/456

### DIFF
--- a/org.osgi.util.promise/src/org/osgi/util/promise/Promise.java
+++ b/org.osgi.util.promise/src/org/osgi/util/promise/Promise.java
@@ -503,4 +503,167 @@ public interface Promise<T> {
 	 * @since 1.2
 	 */
 	CompletionStage<T> toCompletionStage();
+
+	/**
+	 * Register a callback to be called with the failure for this Promise when
+	 * this Promise is resolved with a failure of a failure type. The callback
+	 * will not be called if this Promise is resolved successfully or if the
+	 * failure is not an instance of the specified failure type.
+	 * <p>
+	 * This method may be called at any time including before and after this
+	 * Promise has been resolved.
+	 * <p>
+	 * Resolving this Promise <i>happens-before</i> any registered callback is
+	 * called. That is, in a registered callback, {@link #isDone()} must return
+	 * {@code true} and {@link #getValue()} and {@link #getFailure()} must not
+	 * block.
+	 * <p>
+	 * A callback may be called on a different thread than the thread which
+	 * registered the callback. So the callback must be thread safe but can rely
+	 * upon that the registration of the callback <i>happens-before</i> the
+	 * registered callback is called.
+	 * 
+	 * @param failure The Consumer callback that receives the failure of this
+	 *            Promise if the failure is an instance of the specified failure
+	 *            type. Must not be {@code null}.
+	 * @param failureType The type of failure for which this callback will be
+	 *            called. If the failure is not an instance of the specified
+	 *            failure type, the specified callback will not be called. Must
+	 *            not be {@code null}.
+	 * @param <F> The failure type.
+	 * @return This Promise.
+	 * @since 1.3
+	 */
+	<F> Promise<T> onFailure(Consumer< ? super F> failure,
+			Class< ? extends F> failureType);
+
+	/**
+	 * Recover from a failure of this Promise with a recovery value if the
+	 * failure is an instance of a failure type.
+	 * <p>
+	 * If this Promise is successfully resolved, the returned Promise must be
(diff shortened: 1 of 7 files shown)


org.osgi.util.promise/src/org/osgi/util/promise/Promise.java
```diff
@@ -503,4 +503,166 @@
...
+	 *            Promise if the failure is an instance of the specified failure
+	 *            type. Must not be {@code null}.
+	 * @param failureType The type of failure for which this callback will be
+	 *            called. If the failure is not an instance of the specified
+	 *            failure type, the specified callback will not be called. Must
+	 *            not be {@code null}.
+	 * @return This Promise.
+	 * @since 1.3
+	 */
+	Promise<T> onFailure(Consumer< ? super Throwable> failure,
```
glimmerveen, line 536: Introducing a dependency between the failureType and the failure consumer as follows, would eliminate the need to cast the generic throwable to the failure type within the consumer lambda in case the consumer requires to access any of the failureType's specific methods/fields.
```java
<F extends Throwable> Promise<T> onFailure(Consumer<? super F> failure, Class<F> failureType)
```

org.osgi.util.promise/src/org/osgi/util/promise/Promise.java
```diff
@@ -503,4 +503,166 @@
...
+	 *            type, the specified recovery will not be called and the
+	 *            returned Promise must be failed with the failure of this
+	 *            Promise. Must not be {@code null}.
+	 * @return A Promise that resolves with the value of this Promise or
+	 *         recovers from the failure of this Promise if the failure is an
+	 *         instance of the specified failure type.
+	 * @since 1.3
+	 */
+	Promise<T> recover(Function<Promise< ? >, ? extends T> recovery,
+			Class< ? > failureType);
```
glimmerveen, line 586: Constraining the failureType's type to `Class< ? extends Throwable >` would prevent non-exception types to be passed as failure type.

org.osgi.util.promise/src/org/osgi/util/promise/Promise.java
bjhargrave: `failureType` does not need to be a `Throwable`. It can be an interface which some set of custom exceptions implement. This is why it is `Class<?>` and not `Class<? extends Throwable>`. So we cannot establish a type linkage between the two parameters.

timothyjward: I see why the class might not be a `Throwable`, but surely the consumer can be `? super X` to save the user casting every time?

bjhargrave: No, because F is unbound. There is no type-safe way to guarantee that the exception passed to the consumer can be cast to F's reification. 

`onFailure((RuntimeException e) -> e.printStackTrace(), Throwable.class);`

would be acceptable to the compiler, but at runtime any exception can be passed to the consumer which would fail when the (compiler generated) cast to RuntimeException fails.

bjhargrave: `<F> Promise<T> onFailure(Consumer<? super F> failure,	Class<? extends F> failureType);`

would be a possible signature choice since `F` is then bound by `failureType` and we can be confident that the exception can be passed to the consumer in a type safe manner.

But this only helps `onFailure` which directly consumes the exception. The other methods all get passed a Promise which supplies the exception as a Throwable and would need casting to any other exception type.

timothyjward: > But this only helps onFailure which directly consumes the exception. The other methods all get passed a Promise which supplies the exception as a Throwable and would need casting to any other exception type.

Agreed, but I think it would be good to help here as we can.

glimmerveen: > `failureType` does not need to be a `Throwable`. It can be an interface which some set of custom exceptions implement. This is why it is `Class<?>` and not `Class<? extends Throwable>`. So we cannot establish a type linkage between the two parameters.

I did not consider that you might want to match on an interface implemented by the raised failure :). Perhaps this is an additional use-case to cover in the tests?

bjhargrave: > Perhaps this is an additional use-case to cover in the tests?

Good idea. Done!

Issue #455, opened 2022-06-28
[proposal][jpa] support JNDI Datasources


laeubi:
In the `persistence.xml` it is possible to [define JNDI datasources to use](https://docs.oracle.com/cd/E16439_01/doc.1013/e13981/cfgdepds005.htm#BEIJJCGH).
The OSGi JPA Spec currently only covers the case where a `DataSourceFactory` is used, but especially when running inside managed containers it would be useful to support JNDI as well.

[PAX JPA implementation](https://github.com/ops4j/org.ops4j.pax.jpa/) already has (non standard) support for JNDI datasources in the following way:

1. it first checks if (depending on the transaction type) JtaDataSource or NonJtaDataSource is given in `persistence.xml`
2. if yes, [use a thin wrapper](https://github.com/ops4j/org.ops4j.pax.jpa/blob/master/pax-jpa/src/main/java/org/ops4j/pax/jpa/impl/JndiDataSourceFactory.java) that delegates to JNDI
3. if no proceed as usual

bjhargrave:
When we update the JPA spec, we also need to update for jakarta namespace.

laeubi:
> When we update the JPA spec, we also need to update for jakarta namespace.

Should this be an preliminary step or should it be done as part of this change?

laeubi:
Still an issue

PR #473, opened 2022-09-13
[proposal][jpa] support JNDI Datasources

laeubi:
This aims to fix https://github.com/osgi/osgi/issues/455

### DIFF
--- a/osgi.specs/docbook/127/service.jpa.xml
+++ b/osgi.specs/docbook/127/service.jpa.xml
@@ -1317,6 +1317,26 @@ suffix:     META-INF/persistence.xml</programlisting>
       implementation will need to be included with the JPA Provider
       bundle.</para>
     </section>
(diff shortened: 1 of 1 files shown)


bjhargrave (commented):
This change will necessitate a minor version bump in the org.osgi.service.jpa package (which will the bump the spec version). 

When changing a spec, we also need to add a Changes section to the bottom of the file. See https://github.com/osgi/osgi/blob/d0b971ddb385b7d69b5b44e285336a2292d93fa3/osgi.specs/docbook/705/util.promise.xml#L830-L848 for an example.

Also, a spell check is needed.

laeubi:
@bjhargrave thanks for the guidance, I think one also needs to adjust the TCK?
This PR is really just "draft" state I'll try to enhance this step-by-step.

PR #494, opened 2022-10-26
adding new ThresholdPushbackPolicy

### LINKED ISSUE #442: Additional default PushBackPolicy
In `org.osgi.util.pushstream.PushbackPolicyOption` we have a couple of usefull default policies. Aver the last years, we found another one that became our default and I suggest adding it to the exsisting List:

```
	/**
	 * This {@link PushbackPolicy} increases linear after the queue breaches a given threshold.  
	 * If the threshold is e.g. 100 the 101st element would cause a backpressure of one and the 110th element a pushback of 10. 
	 */
	LINEAR_AFTER_THRESHOLD {
		@Override
		public <T, U extends BlockingQueue<PushEvent<? extends T>>> PushbackPolicy<T, U> getPolicy(long value) {
			return q -> Math.max(0, q.size() - value);
		}
	};
```

juergen-albert:
closes #442

Signed-off-by: Juergen Albert <j.albert@data-in-motion.biz>

### DIFF
--- a/org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
+++ b/org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
@@ -1,0 +1,136 @@
+/*******************************************************************************
+ * Copyright (c) Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0 
+ *******************************************************************************/
+package org.osgi.util.pushstream;
+
+import java.util.concurrent.BlockingQueue;
+
+/**
+ * Provides a configurable {@link PushbackPolicy} implementation that returns
+ * zero back pressure until the buffer fills beyond the supplied threshold. Once
+ * the threshold is reached back pressure is returned based on the supplied
+ * parameters.
+ * <p>
+ * The starting level of the back pressure once the threshold is exceeded is
+ * defined using the <code>initial</code> parameter. Additional back pressure is
+ * applied for each queued item over the threshold. The amount of this
+ * additional back pressure is defined by the <code>increment</code> parameter.
+ * <p>
+ * The following common use cases are supported:
+ * <ul>
+ * <li>The increment size can be zero, returning the initial value as a fixed
+ * back pressure after the threshold is exceeded.</li>
+ * <li>The initial value can be zero, returning a linearly increasing back
+ * pressure from zero once the threshold is exceeded</li>
+ * </ul>
+ * 
+ * @param <T> The type of objects in the {@link PushEvent}
+ * @param <U> The type of the Queue used in the user specified buffer
+ * @since 1.1
+ */
+public final class ThresholdPushbackPolicy<T, U extends BlockingQueue<PushEvent< ? extends T>>>
+		implements PushbackPolicy<T,U> {
+
+	private final int	threshold;
+	private final long	initial;
+	private final long	increment;
+
+	/**
+	 * A simple configuration with an initial back pressure of one and and
+	 * increase increment size of one.
+	 * 
+	 * @see #createThresholdPushbackPolicy(int threshold, long initial, long
+	 *      increment)
+	 * @param threshold the queue size limit after which back pressure will be
+	 *            applied
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 * @return a new {@link ThresholdPushbackPolicy}
+	 */
+	public static <T, U extends BlockingQueue<PushEvent< ? extends T>>> ThresholdPushbackPolicy<T,U> createSimpleThresholdPushbackPolicy(
+			int threshold) {
+		return new ThresholdPushbackPolicy<>(threshold, 1L, 1L);
+	}
+
+	/**
+	 * A simple configuration, where the increment size is used as the initial
+	 * back pressure.
+	 * 
+	 * @see #createThresholdPushbackPolicy(int threshold, long initial, long
+	 *      increment)
+	 * @param increment the increments in which the pressure increases
+	 * @param threshold the queue size limit after which back pressure will be
+	 *            applied
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 * @return a new {@link ThresholdPushbackPolicy}
+	 */
+	public static <T, U extends BlockingQueue<PushEvent< ? extends T>>> ThresholdPushbackPolicy<T,U> createIncrementalThresholdPushbackPolicy(
+			int threshold, long increment) {
+		return new ThresholdPushbackPolicy<>(threshold, increment, increment);
+	}
+
+	/**
+	 * Provides a {@link ThresholdPushbackPolicy} with an individual
+	 * configuration for all possible parameters.
+	 * 
+	 * @param increment the increments in which the pressure increases.
+	 * @param initial the initial back pressure which defines the floor after
+	 *            the threshold is exceeded
+	 * @param threshold the queue size limit after which back pressure will be
+	 *            applied
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 * @return a new {@link ThresholdPushbackPolicy}
+	 */
+	public static <T, U extends BlockingQueue<PushEvent< ? extends T>>> ThresholdPushbackPolicy<T,U> createThresholdPushbackPolicy(
+			int threshold, long initial, long increment) {
+		return new ThresholdPushbackPolicy<>(threshold, initial, increment);
+	}
+	
+	/**
+	 * @param increment the increments in which the pressure increases.
+	 * @param initial the initial back pressure which defines the floor after
+	 *            the threshold is exceeded
+	 * @param threshold the queue size limit after which back pressure will be
+	 *            applied
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 */
+	private ThresholdPushbackPolicy(int threshold, long initial,
+			long increment) {
+		if (threshold < 0 || initial < 0L || increment < 0L) {
+			throw new IllegalArgumentException(String.format(
+					"threshold, initial and increment must each be zero or higher. Current values are threshold [%s], initial [%s], increment [%s]",
+					Integer.valueOf(threshold), Long.valueOf(initial),
+					Long.valueOf(increment)));
+		}
+		this.threshold = threshold;
+		this.initial = initial;
+		this.increment = increment;
+	}
+
+	@Override
+	public long pushback(U queue) throws Exception {
+		int factor = queue.size() - threshold - 1;
+		if (factor < 0) {
+			return 0L;
+		}
+		return initial + (factor * increment);
+	}
+
+}
--- a/org.osgi.util.pushstream/src/org/osgi/util/pushstream/package-info.java
+++ b/org.osgi.util.pushstream/src/org/osgi/util/pushstream/package-info.java
@@ -17,25 +17,23 @@
  *******************************************************************************/
 
 /**
- * Push Stream Package Version 1.0.
- * 
+ * Push Stream Package Version 1.1.
  * <p>
  * Bundles wishing to use this package must list the package in the
  * Import-Package header of the bundle's manifest.
- * 
  * <p>
  * Example import for consumers using the API in this package:
  * <p>
- * {@code Import-Package: org.osgi.util.pushstream; version="[1.0,2.0)"}
+ * {@code Import-Package: org.osgi.util.pushstream; version="[1.1,2.0)"}
  * <p>
  * Example import for providers implementing the API in this package:
  * <p>
- * {@code Import-Package: org.osgi.util.pushstream; version="[1.0,1.1)"}
+ * {@code Import-Package: org.osgi.util.pushstream; version="[1.1,1.2)"}
  * 
  * @author $Id$
  */
 
-@Version("1.0.2")
+@Version("1.1.0")
 package org.osgi.util.pushstream;
 
 import org.osgi.annotation.versioning.Version;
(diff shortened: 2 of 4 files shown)


org.osgi.util.pushstream/src/org/osgi/util/pushstream/package-info.java
```diff
@@ -35,7 +33,7 @@
  * @author $Id$
  */
 
-@Version("1.0.2")
+@Version("1.1.0")
```
bjhargrave, line 36: You also need to update the comment above to use 1.1 instead of 1.0.

osgi.specs/docbook/706/util.pushstream.xml
```diff
@@ -1076,6 +1076,13 @@
         xrefstyle="hyperlink"/> provides basic implementations of the push
         back policies, but custom polices can be implemented to provide more
         complex behaviors.</para>
+        
+        <para>The <xref
+        linkend="org.osgi.util.pushstream.ThresholdPushbackPolicy"
+        xrefstyle="hyperlink"/> provides a specific implementations of a push
+        policy. It covers a common case were back pressure needs to be applied 
+        after a certain threshold in the buffer is exceeded. The intial pressure 
+        and subsequent increase can be configured as well.</para>
```
bjhargrave, line 1085:
```suggestion
        <para>The <xref
        linkend="org.osgi.util.pushstream.ThresholdPushbackPolicy"
        xrefstyle="hyperlink"/> provides a specific implementation of a push
        policy. It covers a common case were back pressure needs to be applied 
        after a certain threshold in the buffer is exceeded. The initial pressure 
        and subsequent increase can be configured as well.</para>
```

osgi.specs/docbook/706/util.pushstream.xml
```diff
@@ -1553,4 +1560,16 @@ psp.createStream(ses).
...
   </section>
+  
+  <section>
+    <title>Changes</title>
+
+    <itemizedlist>
+      <listitem>
+        <para>Added new push back policy implementation <xref
+        linkend="org.osgi.util.pushstream.ThresholdPushbackPolicy"
+        xrefstyle="hyperlink"/></para>
```
bjhargrave, line 1571:
```suggestion
        <para>Added new push back policy implementation <xref
        linkend="org.osgi.util.pushstream.ThresholdPushbackPolicy"
        xrefstyle="hyperlink"/>.</para>
```

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+ * factoring the excess elements and multiplying them with the configured step
+ * added to the initial value.
+ * <p/>
+ * The step size can be zero, which makes it return the initial value as a fixed
+ * back pressure after the threshold is exceeded.
+ * 
+ * @param <T> The type of objects in the {@link PushEvent}
+ * @param <U> The type of the Queue used in the user specified buffer
+ */
+public class ThresholdPushbackPolicy<T, U extends BlockingQueue<PushEvent< ? extends T>>>
```
bjhargrave, line 37:
```suggestion
public final class ThresholdPushbackPolicy<T, U extends BlockingQueue<PushEvent< ? extends T>>>
```

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+	 * @param step the steps in which the pressure increases.
+	 * @param initial the initial back pressure which defines the floor after
+	 *            the threshold is exceeded
+	 * @param threshold the queue size limit after which back pressure will be
+	 *            applied
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 */
+	public ThresholdPushbackPolicy(int threshold, long initial, long step) {
+		if (threshold < 0 || initial < 0 || step < 0) {
```
bjhargrave, line 81:
```suggestion
		if (threshold < 0 || initial < 0L || step < 0L) {
```

We should use proper long constants.

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+	 * A simple configuration with an initial back pressure of one and and
+	 * increase step size of one
+	 * 
+	 * @param threshold the queue size limit after which back pressure will be
+	 *            applied
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 */
+	public ThresholdPushbackPolicy(int threshold) {
+		this(threshold, 1, 1);
```
bjhargrave, line 54:
```suggestion
		this(threshold, 1L, 1L);
```

Proper long constants.

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 */
+	public ThresholdPushbackPolicy(int threshold, long initial, long step) {
+		if (threshold < 0 || initial < 0 || step < 0) {
+			throw new IllegalArgumentException(String.format(
+					"threshold, initial and step muste be larger then zero. Current values are threshold [%s], initial [%s], step [%s]",
+					Integer.valueOf(threshold), Long.valueOf(initial),
+					Long.valueOf(step)));
+		}
```
bjhargrave, line 86:
```suggestion
			throw new IllegalArgumentException(String.format(
					"threshold, initial and step must each be larger than zero. Current values are threshold [%s], initial [%s], step [%s]",
					Integer.valueOf(threshold), Long.valueOf(initial),
					Long.valueOf(step)));
		}
```

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+	/**
+	 * A simple configuration with an initial back pressure of one and and
+	 * increase step size of one
+	 * 
+	 * @param threshold the queue size limit after which back pressure will be
+	 *            applied
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 */
+	public ThresholdPushbackPolicy(int threshold) {
```
bjhargrave, line 53: I thought we agreed to use static methods rather than public constructors.

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+		this.threshold = threshold;
+		this.initial = initial;
+		this.step = step;
+	}
+
+	@Override
+	public long pushback(U queue) throws Exception {
+		int size = queue.size();
+		if (size <= threshold) {
+			return 0;
```
bjhargrave, line 96:
```suggestion
			return 0L;
```

bjhargrave (changes requested):

org.osgi.util.pushstream/src/org/osgi/util/pushstream/package-info.java
juergen-albert: done

osgi.specs/docbook/706/util.pushstream.xml
juergen-albert: done

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
juergen-albert: done

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
juergen-albert: done

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
juergen-albert: done

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
juergen-albert: done

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
juergen-albert: Done. I hope I picked some sensible names for the different methods

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
juergen-albert: done

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0 
+ *******************************************************************************/
+package org.osgi.util.pushstream;
+
+import java.util.concurrent.BlockingQueue;
+
+/**
+ * Provides a configurable {@link PushbackPolicy} implementations. It provides
```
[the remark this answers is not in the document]
juergen-albert: thanks, used your version

osgi.specs/docbook/706/util.pushstream.xml
```diff
@@ -1076,6 +1076,13 @@
         xrefstyle="hyperlink"/> provides basic implementations of the push
         back policies, but custom polices can be implemented to provide more
         complex behaviors.</para>
+        
+        <para>The <xref
+        linkend="org.osgi.util.pushstream.ThresholdPushbackPolicy"
+        xrefstyle="hyperlink"/> provides a specific implementations of a push
```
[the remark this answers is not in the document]
juergen-albert: thanks, used your version

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,133 @@
...
+ * <ul>
+ * <li>The increment size can be zero, returning the initial value as a fixed
+ * back pressure after the threshold is exceeded.</li>
+ * <li>The initial value can be zero, returning a linearly increasing back
+ * pressure from zero once the threshold is exceeded</li>
+ * </ul>
+ * 
+ * @param <T> The type of objects in the {@link PushEvent}
+ * @param <U> The type of the Queue used in the user specified buffer
+ */
```
bjhargrave, line 43:
```suggestion
 * @since 1.1
 */
```

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,133 @@
...
+	 * @param threshold the queue size limit after which back pressure will be
+	 *            applied
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 */
+	private ThresholdPushbackPolicy(int threshold, long initial,
+			long increment) {
+		if (threshold < 0 || initial < 0L || increment < 0L) {
+			throw new IllegalArgumentException(String.format(
+					"threshold, initial and increment must each be larger than zero. Current values are threshold [%s], initial [%s], increment [%s]",
```
bjhargrave, line 114: The code requires non-negative values, but the exception messages says you must use positive values. This needs to be reconciled. My guess is you really want positive values, so the code test must be `<= 0'`.

The new TCK tests should also confirm that 0 is illegal to match this code.

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
```diff
@@ -0,0 +1,133 @@
...
+	 * @throws IllegalArgumentException if any of the given values is lower then
+	 *             zero
+	 * @return a new {@link ThresholdPushbackPolicy}
+	 */
+	public static <T, U extends BlockingQueue<PushEvent< ? extends T>>> ThresholdPushbackPolicy<T,U> createIncrementalThresholdPushbackPolicy(
+			int threshold, long increment) {
+		return new ThresholdPushbackPolicy<>(threshold, increment, increment);
+	}
+
+	/**
```
bjhargrave, line 86: Need comment before param.

bjhargrave (changes requested):

org.osgi.util.pushstream/src/org/osgi/util/pushstream/ThresholdPushbackPolicy.java
juergen-albert: 0 for any or all of its values is actually legal and having one or more of them as such might even make sense in some situation. Thus I allowed for it. One might argue that e.g. all initialized with 0 does not make sense, as it would effectively do nothing, but it is still legal and work fine.

timothyjward: > The new TCK tests should also confirm that 0 is illegal to match this code.

This should not be the case, at least for `initial` and `increment`. The JavaDoc for this class outlines uses `0` for both of these (not simultaneously) to get useful behaviours. Specifically a zero increment gives a “threshold fixed” policy, and a zero initial gives a “threshold linear” policy. 

As Jürgen points out, zero for both initial and increment is a bit strange but not actively broken. Zero for threshold collapses to a combination of linear plus fixed, and might be considered odd enough to be broken.

bjhargrave: OK, then the exception message is wrong since it states the values "must each be larger than zero".

juergen-albert: That is correct. Changed it accordingly.

bjhargrave (approved):

bjhargrave:
OK, but the TCK for org.osgi.service.log.stream failed. Did you diagnose that?

juergen-albert:
Sorry, I didn't catch that. I came to the same conclusion as you did, regarding the log implementation. How do we want to precede? I guess we wait until the PR is merged and released and then merge again?

bjhargrave:
This PR is done. We just need to change to the updated Eclipse Log Stream impl when it is available. Until then, the Log Stream TCK is broken.

Issue #474, opened 2022-09-14
Need CM compatible implementation update for TCK change


bjhargrave:
Related Issue in Felix-Jira: https://issues.apache.org/jira/browse/FELIX-6470

_Originally posted by @stbischof in https://github.com/osgi/osgi/issues/343#issuecomment-1163116780_

cziegeler:
This case actually raises an interesting question about how notifications happen when the coordination finishes.
Simplified use case: managed service is registered, it gets updated with null. Now a coordination starts, a configuration is created, and then deleted before the coordination finishes. The coordination then finishes. Is the managed service notified at all as the state before the coordination and after the coordination is the same?
The specification is open about this, but my interpretation is that it is ok for the CM implementation to not send notifications in such a case, as they would only cause unnecessary churn.
If we follow that thought, then we could say that the notifications that are send out after the coordination finishes "report" the diff between the state before the coordination started and after. For example, if a configuration is updated several times throughout a coordination, only one notification is send out with the latest configuration.

bjhargrave:
For coordinations, the goal is to avoid observing intermediate states. So when the coordination ends, CM must only notify with the final config state. So this means the CM impl should be able to collapse multiple changes to a config during a coordination to a single notification. 

Given this, both the TCK tests for coordination and the impl need to change.

stbischof:
I take the tck part.

cziegeler:
@stbischof  I just updated the implementation in Apache Felix. There, I added these tests: https://github.com/apache/felix-dev/blob/master/configadmin/src/test/java/org/apache/felix/cm/integration/CoordinatorIntegrationTest.java

PR #498, opened 2022-10-27
Update CM TCK tests for coordination

stbischof:
ref: #474


cziegeler (approved):

cziegeler:
@stbischof Thanks for updating the tests; I think those two tests are good with the proposed changes. However, we also need to adjust the existing tests:
- testCoordinatedConfigurationOnBeforeRegisteredManagedService is currently checking for four events to happen; however with the proposed changes, it will just be one (the first one invoking the managed service with null). During the coordination the configuration is created, updated and then deleted, resulting in no invocation of the managed service
- testManagedService - similar here, the configuration is created, updated, deleted - managed service will only be invoked with "null" once
I have the feeling that we then have an overlap between those two new tests and the two that need adjustment; but that's probably ok.

cziegeler:
I've created #505 which superceeds (includes) this PR

PR #515, opened 2022-11-23
Servlet test cases must not expect default SCH in service registry

cziegeler:
Signed-off-by: Carsten Ziegeler <cziegeler@apache.org>


cziegeler:
The updated test cases reveal a bug in the Apache Felix Http Jetty implementation (getMimeType() not returning null for the default context). This bug is fixed in version 5.0.0-RC2 which is currently under vote and will be released end of week.

bjhargrave (approved):

Issue #530, opened 2022-12-21
JDBCTestCase.testMinimumOneDataSourceFactory(..) is not parameterized


laeubi:
org.osgi.test.cases.jdbc.junit.JDBCTestCase.testMinimumOneDataSourceFactory(ServiceAware<DataSourceFactory>) is the only method that is not parameterized

PR #531, opened 2022-12-21
JDBCTestCase.testMinimumOneDataSourceFactory(..) should be parameterized

laeubi:
@bjhargrave @stbischof can you please review?

This changes two things:

1. Make the test parameterized and inspect any jdbc service (as with the other methods in this test)
2. remove requirement that the service can be materialized. The testname and the spec only mandate that the service is registered with this properties, I don't think this needs to be asserted here that the service actually exits as other test methods will ensure this.

Fix https://github.com/osgi/osgi/issues/530

### DIFF
--- a/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
+++ b/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
@@ -37,8 +37,6 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.osgi.framework.ServiceReference;
 import org.osgi.service.jdbc.DataSourceFactory;
 import org.osgi.test.assertj.servicereference.ServiceReferenceAssert;
-import org.osgi.test.common.annotation.InjectService;
-import org.osgi.test.common.service.ServiceAware;
 import org.osgi.test.junit5.service.ServiceSource;
 
 public class JDBCTestCase {
@@ -77,20 +75,21 @@ public class JDBCTestCase {
 	static String				description										= "desc";
 	static String				password										= "pswd";
 	static String				user											= "usr";
-
-	@Test
+	
+	/**
+	 * Test that the JDBS service defines at least one capability
+	 * 
+	 * @param sr the service reference parameter
+	 */
+	@ParameterizedTest
+	@ServiceSource(serviceType = DataSourceFactory.class)
 	public void testMinimumOneDataSourceFactory(
-			@InjectService(cardinality = 1, filter = "("
-					+ OSGI_JDBC_DRIVER_CLASS + "=*)")
-			ServiceAware<DataSourceFactory> sa) {
-		assertThat(sa.getService()).isNotNull();
-
-		ServiceReferenceAssert.assertThat(sa.getServiceReference())
+			ServiceReference<DataSourceFactory> sr) {
+		ServiceReferenceAssert.assertThat(sr)
 				.hasServicePropertiesThat()
 				.containsKey(OSGI_JDBC_CAPABILITY)
 				.extractingByKey(OSGI_JDBC_CAPABILITY)
 				.isNotNull();
-
 	}
 
 	@ParameterizedTest


org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -34,9 +34,11 @@
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceReference;
 import org.osgi.service.jdbc.DataSourceFactory;
 import org.osgi.test.assertj.servicereference.ServiceReferenceAssert;
+import org.osgi.test.common.annotation.InjectBundleContext;
```
bjhargrave, line 41: The new imports seem extraneous. Also the imports for InjectService and ServiceAware are now unused. Please organize imports.

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -77,20 +79,21 @@ private static class DfltProps extends Properties {
...
-
-	@Test
+	
+	/**
+	 * Test that the JDBS service defines at least one capability
+	 * 
+	 * @param sr the service reference parameter
+	 */
+	@ParameterizedTest
+	@ServiceSource(serviceType = DataSourceFactory.class)
```
bjhargrave, line 89: You have lost the filter expression `"(" + OSGI_JDBC_DRIVER_CLASS + "=*)"` from the test.

bjhargrave (changes requested):

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
laeubi: This seems redundant here and is tested in `testRegisteredProperties` already.

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
laeubi: This should be fixed now!

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
bjhargrave: I think the point of this test method is to confirm there is at least one DSF with the driver class property and the capability. Which is why it was not a parameterized test. I am now thinking we should leave this test method alone. Making it parameterized does not add any value and seems to lose some of the purpose of the test.

laeubi: This makes no sense to me... `OSGI_JDBC_DRIVER_CLASS` is a required property, that means every service must implement it to be compliant. And only testing one here but multiple ones in the other methods seems really odd, so my implementation becomes compliant to the TCK if I register multiple ones where only some has the required properties?

laeubi: By the way does `@InjectService(cardinality = 1)` mean *exactly* one or *at laest* one? So it would be good to clarify if this should test exactly one service or several ones. If only one I still wonder why a filter is required and not the service type is enough?

And then the other methods should also simply collect one service instead of potentially multiple ones...

laeubi: The spec says:

> Each Data Source Factory service must relate to a single Database Driver. The Database Driver implementation bundle does not necessarily need to be the registrar of the Data Source Factory service. Any bundle can provide the Data Source Factory service and delegate to the appropriate driver specific implementation classes. 

This implies to me a compliant implementation can implement as many `DataSourceFactory`s as desired (it is unclear what "must relate to a single Database Driver" should mean in the context of the spec), so each of those factories must comply to any rules, and it is even invalid to check this in different methods as there is no guarantee the same number of services are checked.

So I think actually the checks that are now separate testing for properties are not compliant to the spec ... if you agree I'll therefore rework both PRs to take this into account.

laeubi:
I'm closing this as it needs a different approach...

Issue #529, opened 2022-12-21
[jdbc][tck] JDBCTestCase fails if driver do not implement all optional methods


laeubi:
https://github.com/osgi/osgi/blob/main/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java currently fails with an obscure error (see https://github.com/osgi/osgi-test/issues/593) if the datasource do not implement all properties, but the spec says that this is fine:

> A Data Source Factory service is not required to implement all of the factory methods. To indicate which factory method is implemented, the Data Source Factory service must be registered with the service property osgi.jdbc.datasourcefactory.capability having one or more of the following values.
> 
> -  driver
> -  datasource
> -  connectionpooldatasource
> -  xadatasource

Also I think the following is not covered
> If an implementation does not support a particular type then it must throw a SQL Exception. 

the TCK only checks that no exception is thrown, but not if exception is thrown otherwhise.

PR #532, opened 2022-12-21
JDBCTestCase fails if driver do not implement all optional methods

laeubi:
This also includes
- https://github.com/osgi/osgi/pull/531

an fixes https://github.com/osgi/osgi/issues/529

@bjhargrave @stbischof can you please review?

This changes

- always collect all services
- check if the service claims to implement the optional method and assert it creates the expsected items
- if not, assert that it throw the expected SQLExceptions

### DIFF
--- a/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
+++ b/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
@@ -34,9 +34,11 @@ import java.util.Properties;
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceReference;
 import org.osgi.service.jdbc.DataSourceFactory;
 import org.osgi.test.assertj.servicereference.ServiceReferenceAssert;
+import org.osgi.test.common.annotation.InjectBundleContext;
 import org.osgi.test.common.annotation.InjectService;
 import org.osgi.test.common.service.ServiceAware;
 import org.osgi.test.junit5.service.ServiceSource;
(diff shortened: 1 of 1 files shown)


org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -77,20 +84,24 @@ private static class DfltProps extends Properties {
...
+	@InjectBundleContext
+	BundleContext bundleContext;
+	
+	/**
+	 * Test that the JDBS service defines at least one capability
+	 * 
+	 * @param sr the service reference parameter
+	 */
+	@ParameterizedTest
+	@ServiceSource(serviceType = DataSourceFactory.class)
```
bjhargrave, line 97: Doesn't this PR contain changes which are the same as the other PR? We should merge the other PR first and then rebase this PR.

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -31,12 +32,18 @@
 
 import java.sql.SQLException;
 import java.util.Properties;
+import java.util.function.BiFunction;
```
bjhargrave, line 35: I think this is unneeded. Perhaps organize imports.

bjhargrave (changes requested):

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
laeubi: Yep see description, the other should be merged first and I'll then rebase it.

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -118,66 +129,57 @@ public void testRegisteredProperties(
...
+				assertThatNoException().isThrownBy(() -> function.invoke(dsf, new DfltProps()));
+				assertThat(function.invoke(dsf, new DfltProps())).isNotNull();
+			} else {
+				//it is required to throw an exception here
+				assertThatExceptionOfType(SQLException.class).isThrownBy(() -> function.invoke(dsf, null));
+				assertThatExceptionOfType(SQLException.class).isThrownBy(() -> function.invoke(dsf, new DfltProps()));
+			}
+		} finally {
+			if (dsf != null) {
+				bundleContext.ungetService(reference);
```
stbischof, line 176: Not sure if this is needes. The injected BundleContext is [CloseableBundleContext](https://github.com/osgi/osgi-test/blob/main/org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java).

laeubi: Hard to guess, there is not a single line of documentation at that class, but it seems not wrong to be nice and cleanup ourselfs :-)

laeubi:
I'm closing this as it needs a different approach...

PR #537, opened 2022-12-22
Rework the JDBC TCK

laeubi:
This do the following:
- merge the method testRegisteredProperties / testMinimumOneDataSourceFactory as it is not compliant to only test there is one valid DSF but all need to implement all required properties
- check that the description (if given) is a string property
- check that the capability is either a string or a String[] type
- join the individual testCapabilityX into one method as actually one needs to check all declared capabilities of one source fulfills the overall contract of at least provide one.

Fix https://github.com/osgi/osgi/issues/529
Fix https://github.com/osgi/osgi/issues/530

@bjhargrave @stbischof please have a look. This should now be consistent to the spec and could test any number of DSF registered.

Additionally this now also check that a DSF at least implements one of the valid values (in contrast to before where it was only checked that one DSF declares any value) and also checks that the property if of the desired type `String+`

### DIFF
--- a/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
+++ b/org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
@@ -19,29 +19,46 @@
 package org.osgi.test.cases.jdbc.junit;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatNoException;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_CAPABILITY;
 import static org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_CAPABILITY_CONNECTIONPOOLDATASOURCE;
 import static org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_CAPABILITY_DATASOURCE;
 import static org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_CAPABILITY_DRIVER;
 import static org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_CAPABILITY_XADATASOURCE;
 import static org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_DRIVER_CLASS;
+import static org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_DRIVER_NAME;
 import static org.osgi.service.jdbc.DataSourceFactory.OSGI_JDBC_DRIVER_VERSION;
 
+import java.sql.Driver;
 import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
 import java.util.Properties;
+import java.util.function.Supplier;
 
-import org.junit.jupiter.api.Test;
+import javax.sql.ConnectionPoolDataSource;
+import javax.sql.DataSource;
+import javax.sql.XADataSource;
+
 import org.junit.jupiter.params.ParameterizedTest;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.InvalidSyntaxException;
 import org.osgi.framework.ServiceReference;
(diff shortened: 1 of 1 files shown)


org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -57,127 +74,155 @@ private static class DfltProps extends Properties {
...
-
-		ServiceReferenceAssert.assertThat(sa.getServiceReference())
-				.hasServicePropertiesThat()
-				.containsKey(OSGI_JDBC_CAPABILITY)
-				.extractingByKey(OSGI_JDBC_CAPABILITY)
-				.isNotNull();
-
-	}
-
+	private static final List<JdbcCapability<?>> ALL_CAPABILITIES = List.of(
```
bjhargrave, line 77: We still need to support Java 8. So we cannot use List.of. I am surprise you do not get a compilation error here.

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -57,127 +74,155 @@ private static class DfltProps extends Properties {
...
 				.isInstanceOf(String.class);
+		
+		sRefAssert.hasServicePropertiesThat()
+				.as(String.format("According to '125.2 Database Driver' the DataSourceFactory must specify the '%s' with its registration.",
+						OSGI_JDBC_CAPABILITY))
+				.containsKey(OSGI_JDBC_CAPABILITY)
+				.extractingByKey(OSGI_JDBC_CAPABILITY)
+				.isNotNull()
+				.as("According to '125.2 Database Driver' the property '%s' must be a String+", OSGI_JDBC_CAPABILITY)
+				.isInstanceOfAny(String.class, String[].class);
```
bjhargrave, line 123: String+ includes a collection of String.

    For example String+, indicates that a String, a String[], and a Collection<String> are all valid forms.

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -57,127 +74,155 @@ private static class DfltProps extends Properties {
...
-		assertThat(dsf.createDataSource(null)).isNotNull();
-
-		assertThatNoException()
-				.isThrownBy(() -> dsf.createDataSource(new DfltProps()));
-
-		assertThat(dsf.createDataSource(new DfltProps())).isNotNull();
+	private <T> boolean testCapability(ServiceReference<DataSourceFactory> reference, JdbcCapability<T> capability) throws SQLException, InvalidSyntaxException {
+		DataSourceFactory dsf = bundleContext.getService(reference);
+		try {
+			if (FrameworkUtil.createFilter(String.format("(%s=%s)", OSGI_JDBC_CAPABILITY, capability.capability)).match(reference)) {
```
bjhargrave, line 162: You have a bundle context here. Use it to create the Filter instead of FrameworkUtil.

org.osgi.test.cases.jdbc/src/org/osgi/test/cases/jdbc/junit/JDBCTestCase.java
```diff
@@ -57,127 +74,155 @@ private static class DfltProps extends Properties {
...
+				.as("According to '125.2 Database Driver' the property '%s' must be a String+", OSGI_JDBC_CAPABILITY)
+				.isInstanceOfAny(String.class, String[].class);
 
+		//then check optional ones
 		Object driverVersion = sr.getProperty(OSGI_JDBC_DRIVER_VERSION);
-
 		if (driverVersion != null) {
 			assertTrue(driverVersion instanceof String,
-					"The DataSourceFactory driver version is not a String");
+					String.format("According to '125.2 Database Driver' the property '%s' must be a String if specified", OSGI_JDBC_DRIVER_VERSION));
```
bjhargrave, line 129: This sort of test would be better as

```java
assertThat(driverVersion).as(...).isInstanceOf(String.class);
```

bjhargrave (changes requested):

laeubi:
I have now applied the suggested changes.

bjhargrave (approved):

bjhargrave:
Thanks for this PR!

stbischof:
Thx @laeubi good job

Issue #546, opened 2023-01-06
Change InvalidSyntaxException to RuntimeException


laeubi:
As suggested by @tjwatson in https://github.com/osgi/osgi/discussions/541#discussioncomment-4603565
and by @bjhargrave in https://github.com/osgi/osgi/discussions/541#discussioncomment-4606314

the `InvalidSyntaxException` could become a `RuntimeException` to prevent requiring handling checked exceptions.

PR #550, opened 2023-01-07
core: Make InvalidSyntaxException a runtime exception

bjhargrave:
This is binary compatible since we are just introducing a new super type whose super type is the previous super type.

We also remove "throws InvalidSyntaxException" from the framework API since it is OSGi policy to not declare runtime exceptions in the throws clause of a method's declaration. We still javadoc the throwing of the exception.

Filter tests are updated to JUnit 5 and osgi-test.

Fixes https://github.com/osgi/osgi/issues/546


bjhargrave:
Note: This will fail the framework API signature test since the current framework compatible implementation does not use the API updated in this PR.

PR #572, opened 2023-02-15
Remove dependency on bcel by directly rewriting class bytes

### LINKED ISSUE #563: [tck]framework] should embedd bcel
`org.osgi.test.cases.framework 8.0.0.202106111751` currently requires `org.apache.bcel.classfile`, it would we good if TCK tests embeds 3rdparty dependencies that are only needed internally instead of importing them.

timothyjward:
The test weaving is sufficiently simple that we can simply search for a series of bytes in the class file and directly replace them.

Fixes #563

### DIFF
--- a/cnf/ext/central.mvn
+++ b/cnf/ext/central.mvn
@@ -174,8 +174,6 @@ log4j:log4j:1.2.15
 
 com.h2database:h2:1.4.193
 
-org.apache.servicemix.bundles:org.apache.servicemix.bundles.bcel:5.2_4
-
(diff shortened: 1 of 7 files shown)


timothyjward:
> That is, git grep bcel should show no use in the source.

The only reference now is in the licenses, where apparently BCEL is repackaged inside Xalan. I didn't remove that as I have no idea whether Xalan is still in use. I did remove the direct references to the BCEL license.

bjhargrave (approved):

tjwatson (approved):

Issue #575, opened 2023-03-27
[proposal] Jakarta WebSocket Whiteboard


tcalmant:
It would be interesting to be able to register WebSocket handlers using a whiteboard pattern, to mimic servlets handling in chapter 140.

The WebSocket handlers would use the Jakarta WebSocket 2.x annotations to be notified of events (`@OnMessage`, `@OnOpen`, ...).
Unlike the Servlet whiteboard, I don't think it is necessary to handle filters, pre-processors nor resources services due to the lack of specification of the payload in messages.

I'm not sure if the registration path of the WebSocket would be defined using the `@ServerEndpoint` Jakarta annotation or through a specific configuration like `@HttpWhiteboardContext` for servlets.
It would also be necessary to define an `@RequireWebsocketWhiteboard` annotation and the matching capabilities.

References:
* [OSGi Compendium R8 - 140 Jakarta Servlet Whiteboard](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.servlet.html)
* [Jakarta Websocket 2.0](https://jakarta.ee/specifications/websocket/2.0/websocket-spec-2.0.html)
* [Jakarta Websocket 2.1](https://jakarta.ee/specifications/websocket/2.1/jakarta-websocket-spec-2.1.html)

tcalmant:
Draft requirements can be found here: https://github.com/tcalmant/osgi/blob/requirements/575/.design/requirements-575.md

grgrzybek:
I remember about this - I was busy with Camel stuff, but soon I can check this again.

stbischof:
I opened the design branch: https://github.com/osgi/osgi/tree/design/575

paulrutter:
Related PR that enables use of Jetty websockets in Felix HTTP. Do note the need for several workarounds to make it work, although thats probably Jetty specific.
https://github.com/apache/felix-dev/pull/298

paulrutter:
I was reading the drafts (especially https://github.com/tcalmant/osgi/blob/requirements/575/.design/requirements-575.md#uc1-define-and-register-a-websocket-server-endpoint-using-annotations) and have a few questions:

- would it be possible to define a pattern instead of a exact path like in the example? Like is possible when registering a servlet to the whiteboard?
- if so, how would the implementing class have access to the current websocket url? Via the Session object?

[...]

PR #598, opened 2023-06-26
Requirements for a Jakarta WebSockets Whiteboard

tcalmant:
Initial version of the requirements for the proposal of Jakarta WebSocket Whiteboard as described in #575

### DIFF
--- a/.design/requirements-575.md
+++ b/.design/requirements-575.md
@@ -1,0 +1,95 @@
+# Jakarta WebSocket Whiteboard
+
+This documents looks into the requirements for a Jakarta WebSocket whiteboard specification.
+
+## Introduction
+
+The OSGi Compendium R6 introduced the HTTP Whiteboard specification, upgraded in R8.1 to the Whiteboard specification for Jakarta Servlet.
+It provides a light and convenient way of using servlets in a an OSGi environment through the use of the Whiteboard Pattern.
+
+WebSocket is communication protocol designed to work over the HTTP protocol.
+It enables bidirectional exchange of messages between the web client and the server.
+Like an HTTP servlet, a WebSocket handler is bound to a URL.
+
+The conversion from an HTTP connection to a WebSocket stream is done by sending a `GET` HTTP request with a set of headers requesting a connection upgrade.
+The server will respond with a *101 Switching protocols* response.
+
+The WebSocket protocol is defined by [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455) and a Java specification is defined in Jakarta EE.
+This document considers the specification will rely on [Jakarta WebSocket 2.1](https://jakarta.ee/specifications/websocket/2.1/), from Jakarta EE 10.
+
+HTTP server implementations can provide support for WebSocket, but their configuration is complex in the context of the OSGi Whiteboard specification for Jakarta Servlet.
+
+A Jakarta Websocket Whiteboard would ease the deployment of WebSocket handlers in an OSGi HTTP environment.
+This document describes the requirements for such a specification.
+
+## Terminology
+
+* _WebSocket_ - The communication channel based on the WebSocket protocol.
+* _WebSocket Handler_ - Server-side handler of a WebSocket frame handler
+
+## Problem Description
+
+Registering a WebSocket handler can currently be done by using a custom configuration of an HTTP servlet.
+The problem is that such configuration depends on the underlying HTTP server implementation.
+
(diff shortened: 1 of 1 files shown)


stbischof:
@tcalmant could you look for your eclipse foundation eca declaration? and you need to sign-off the commits.

@timothyjward you may could assist.

grgrzybek:
For the record, Pax Web 8 implements whiteboard tracker for _web sockets_ and the details are:

* the service filter `(|(org.ops4j.pax.web.http.whiteboard.websocket=true)(websocket=true))` is used [here](https://github.com/ops4j/org.ops4j.pax.web/blob/web-8.0.20/pax-web-extender-whiteboard/src/main/java/org/ops4j/pax/web/extender/whiteboard/internal/tracker/WebSocketTracker.java#L36-L37)
* services of `objectClass=java.lang.Object` are tracked (as any annotated class can be used)
* an `@javax.websocket.server.ServerEndpoint` annotation is taken from the class of service object to get some metadata (like path, decoders/encoders and subprotocols) [here](https://github.com/ops4j/org.ops4j.pax.web/blob/web-8.0.20/pax-web-spi/src/main/java/org/ops4j/pax/web/service/spi/model/elements/WebSocketModel.java#L154-L166)
* the important implementation aspect is that websockets for running jetty/tomcat/undertow can be correctly installed via SCIs, so special [PaxWebWebSocketsServletContainerInitializer](https://github.com/ops4j/org.ops4j.pax.web/blob/web-8.0.20/pax-web-websocket/src/main/java/org/ops4j/pax/web/websocket/internal/PaxWebWebSocketsServletContainerInitializer.java) is used that does the magic
* Pax Web doesn't necessarily instantiate the websocket endpoint - actually registered object is used as is (so the websocket endpoint instance used is the same as the one registered)
* when Class object is registered, `javax.websocket.server.ServerContainer#addEndpoint(java.lang.Class<?>)` is called
* when actual Object is registered, `javax.websocket.server.ServerContainer#addEndpoint(javax.websocket.server.ServerEndpointConfig))` is called that uses existing instance
* Pax Web ensures that the behavior is consistent for Jetty, Tomcat and Undertow
* Here's the integration test: https://github.com/ops4j/org.ops4j.pax.web/blob/web-8.0.20/pax-web-itest/pax-web-itest-container/pax-web-itest-container-common/src/main/java/org/ops4j/pax/web/itest/container/websockets/AbstractWebSocketWhiteBoardIntegrationTest.java

@tcalmant I checked your draft specification and it roughly matches to what is available in Pax Web.
I didn't think about dedicated annotations like `@WebSocketWhiteboardServletPattern` - I simply used the one from JSR 356 (`@javax.websocket.server.ServerEndpoint`)

Also while from JakartaEE point of view, Servlets and WebSockets are different specs, in Pax Web I treated servlets (+filters, listeners, error pages, ...) **and** websockets in exactly the same way. With this in mind I could look at this proposal as kind of "extension" of Http Whiteboard (now Servlet Whiteboard) OSGi CMPN specification. Of course it's nice to have separate chapter. I'd simply emphasize it's _kind of_ like Servlet Whiteboard specification got extra service to track (WebSocket endpoint class or instance).

I definitely don't feel like I should dictate here. Simply if you want, please have a look at how Pax Web implements WebSocket whiteboard-tracking already.

regards
Grzegorz Grzybek

timothyjward:
After today's call it feels like we're moving beyond requirement gathering. Most of the discussion was about the feasibility of implementing the requirements as written.

At the next call we should consider whether we're ready to move to the next stage (i.e. spec and design)

As for implementation, it appears that *most* implementations are container specific, i.e. they cannot be deployed in a generic web container. The [Tyrus Servlet Container](https://github.com/eclipse-ee4j/tyrus/tree/master/containers/servlet/src/main/java/org/glassfish/tyrus/servlet) however appears to be platform agnostic. This could be suitable for a generic websocket whiteboard implementation on top of the servlet whiteboard, implemented using a `Filter` or similar.

Other container specific implementations could be created using the native adapters (e.g. Jetty or Undertow)

grgrzybek:
For Tyrus I did a quick check
* it indeed is a standalone WebSocket server implementation
* `org.glassfish.tyrus.servlet.TyrusServletFilter` is a _bridge_ from web application (web.xml, WAR) into Tyrus
* this filter uses `org.glassfish.tyrus.servlet.TyrusHttpUpgradeHandler` which deals with WebSockets protocol by taking the data directly from `jakarta.servlet.http.WebConnection` (and its in/out streams)
* this filter + upgrade handler are part of `org.glassfish.tyrus:tyrus-container-servlet` Maven module
* the standalone implementation is `org.glassfish.tyrus:tyrus-core`

PR #603, opened 2023-07-07
Hash pin GitHub Actions

### LINKED ISSUE #592: Hash pin actions on github workflows
Hi I'm Joyce from Google and I'm working on helping many open source projects on improving their supply chain security posture.

## Description

I would like to suggest a security practice recommended by the [OpenSSF Scorecard][scorecard-repo] which is to hash pin CI dependencies to prevent from tag renaming attacks. This means hash pinning GitHub Workflow actions.

Although the actions that need to be hash pinned on osgi are Github Owned Actions, it is still recommended to hash pin them since they are open sourced and are exposed to the same threat vectors any open source project is. Besides, this is currently the only way of using github actions as immutable releases.

Also it might be important to notice that the dependabot, that is already enabled, is able to update both the hash and the comment version related to it.

Let me know if you are open to evaluate those changes and I'll submit the PR ASAP. 

Any questions or concerns just let me know.
Thanks!

## Additional Context

A tag renaming attack is a type of attack whereby an attacker:
- Hijack an action.
- Upload a malicious version.
- Replace existing tags with malicious versions.

See [Why you should pin actions by commit hash](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash) for a better understanding of a tag renaming attack.

[scorecard-repo]: https://github.com/ossf/scorecard

joycebrum:
Closes #592 

### Changes

- I've hash pin GitHub Actions by using step security to help on identifying the right hash commit.

### DIFF
--- a/.github/workflows/bnd-test.yml
+++ b/.github/workflows/bnd-test.yml
@@ -53,21 +53,21 @@ jobs:
     if: ${{ github.repository == 'osgi/osgi' }}
     steps:
     - name: Git Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
     - name: Set up Java
-      uses: actions/setup-java@v3
+      uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
       with:
         distribution: 'temurin'
         java-version: ${{ env.java_version }}
     - name: Build
       id: build
-      uses: gradle/gradle-build-action@v2
+      uses: gradle/gradle-build-action@40b6781dcdec2762ad36556682ac74e31030cfe2 # v2.5.1
       with:
         cache-read-only: ${{ github.ref != 'refs/heads/main' }}
         arguments: --continue :buildscriptDependencies :build :osgi.specs:specifications :osgi.tck:tck.core
     - name: Upload TCK Test Reports on failure
       if: ${{ failure() }}
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
       with:
         name: TCK-core-reports
         path: |
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -59,25 +59,25 @@ jobs:
       tck-matrix: ${{ steps.build.outputs.tck-matrix }}
     steps:
     - name: Git Checkout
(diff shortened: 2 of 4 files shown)


timothyjward:
Thank you @joycebrum - we would very much like to accept this change, but it would help to know whether you are able to sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php). If so then this will make accepting the change very easy, if not then we can investigate with the Eclipse IP team whether it's possible to get an exception for this PR.

.github/workflows/bnd-test.yml
```diff
@@ -53,21 +53,21 @@ jobs:
...
     - name: Set up Java
-      uses: actions/setup-java@v3
+      uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
       with:
         distribution: 'temurin'
         java-version: ${{ env.java_version }}
     - name: Build
       id: build
-      uses: gradle/gradle-build-action@v2
+      uses: gradle/gradle-build-action@40b6781dcdec2762ad36556682ac74e31030cfe2 # v2.5.1
```
stbischof, line 64: https://github.com/gradle/gradle-build-action/releases/tag/v2.5.1

.github/workflows/cibuild.yml
```diff
@@ -161,12 +161,12 @@ jobs:
...
     steps:
     - name: Download Specification HTML
-      uses: actions/download-artifact@v3
+      uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
       with:
         name: OSGi-Specification-HTML
         path: osgi.specs/generated/html
     - name: Publish Specification HTML
-      uses: peaceiris/actions-gh-pages@v3
+      uses: peaceiris/actions-gh-pages@373f7f263a76c20808c831209c920827a82a2847 # v3.9.3
```
stbischof, line 169: https://github.com/peaceiris/actions-gh-pages/releases/tag/v3.9.3

stbischof:
I checked all versions and tags.
- wrapper-validation-action/v1.0.5 exists in an newer version but this be picked up by dependabot later

joycebrum:
Sorry for the delay on accepting the CLA. I had some issues regarding the associated email.

I've upgraded the wrapper to 1.0.6 (https://github.com/gradle/wrapper-validation-action/commit/8d49e559aae34d3e0eb16cde532684bc9702762b), nice catch.

Thanks for double checking it @stbischof !

stbischof:
@joycebrum could you please sign off you last commit?

timothyjward:
LGTM

PR #604, opened 2023-07-10
Fix missing wait in the Application Lifecycle tests

timothyjward:
The Jakarta REST Whiteboard is not required to synchronously update when a service is registered, but instead uses a change count to indicate that a change has been made. The ApplicationPath test set up a Promise to wait for this, but it was never actually used, causing intermittent failures with asynchronous updates.


rotty3000 (approved):
LGTM

PR #605, opened 2023-07-10
Fix test failures with asynchronous implementations

timothyjward:
The Jakarta REST Whiteboard is not required to synchronously update when a service is registered, but instead uses a change count to indicate that a change has been made. The test suites assume that the whiteboard will be in a clean and stable state when they start, but this is not the case when the whiteboard updates asynchronously. Specifically the services registered in tests are unregistered at the end of those tests and then the next test starts immediately. The result of this is that the whiteboard may still be processing the service removals during the start of the next test, this in turn messes up the change counting for the next test and causes failures.

This change adds a loop at the end of each test which waits until the whiteboard change count remains stable for a short (200ms) period.


stbischof (approved):
LGTM

Issue #597, opened 2023-06-23
[Servlet Whiteboard] Prepare Servlet Whiteboard API for use of Servlet 6


maho7791:
The imports of the org.osgi.service.servlet API are restricted to:
jakarta.servlet;version="[5.0,6)",
jakarta.servlet.http;version="[5.0,6)"

We should prepare the API for Servlet 6

grgrzybek:
Changes between 5 and 6 (interfaces):
* added `jakarta.servlet.descriptor.JspPropertyGroupDescriptor.getErrorOnELNotFound()`
* deprecated `jakarta.servlet.http.Cookie.set|getComment()`
* deprecated `jakarta.servlet.http.Cookie.set|getVersion()`
* added `jakarta.servlet.http.Cookie.setAttribute()`
* added `jakarta.servlet.http.Cookie.getAttribute()`
* added `jakarta.servlet.http.Cookie.setAttributes()`
* removed `jakarta.servlet.http.HttpServletRequest.isRequestedSessionIdFromUrl()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpServletRequestWrapper.isRequestedSessionIdFromUrl()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpServletResponse.encodeUrl()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpServletResponse.encodeRedirectUrl()` (deprecated in version 5. `encodeRedirectURL()` should be used)
* removed `jakarta.servlet.http.HttpServletResponse.setStatus()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpServletResponseWrapper.encodeUrl()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpServletResponseWrapper.encodeRedirectUrl()` (deprecated in version 5. `encodeRedirectURL()` should be used)
* removed `jakarta.servlet.http.HttpServletResponseWrapper.setStatus()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpSession.getSessionContext()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpSession.getValue()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpSession.getValueNames()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpSession.putValue()` (deprecated in version 5)
* removed `jakarta.servlet.http.HttpSession.removeValue()` (deprecated in version 5)
[...]

PR #611, opened 2023-07-30
Upgrade of `org.osgi.service.servlet` to support Jakarta Servlet API 6.x

ideas-into-software:
To address requirement https://github.com/osgi/osgi/issues/597, also as a prerequisite of upgrading Apache Felix to Jakarta Servlet API 6.x (see https://issues.apache.org/jira/browse/FELIX-6612 and https://github.com/apache/felix-dev/pull/218 for more info).

As per discussion with Mark Hoffmann (Data In Motion), please clarify: 
 - whether `org.osgi.test.cases.servlet` and `org.osgi.test.cases.servlet.secure` should also be upgraded in such way, as they depend on `org.osgi.service.servlet`; 
  - whether `Require-Capability` manifest header of `org.osgi.service.servlet` should be enhanced with `osgi.contract` info for `JakartaServlet`, as currently it only mentions Java version;


ideas-into-software:
Kindly please address questions asked in PR description (https://github.com/osgi/osgi/pull/611#issue-1828103789). The failing checks ("80 successful, 2 failing, and 1 skipped checks") are for `org.osgi.test.cases.servlet` and `org.osgi.test.cases.servlet.secure` and one of my questions specifically mentions those bundles.

org.osgi.service.servlet/src/org/osgi/service/servlet/whiteboard/HttpWhiteboardConstants.java
```diff
@@ -571,5 +571,5 @@ private HttpWhiteboardConstants() {
 	 * The version of the implementation capability for the Servlet Whiteboard
 	 * specification
 	 */
-	public static final String	HTTP_WHITEBOARD_SPECIFICATION_VERSION				= "2.0";
+	public static final String	HTTP_WHITEBOARD_SPECIFICATION_VERSION				= "3.0";
```
stbischof, line 574: 2.1 should be enough.
@cziegeler

cziegeler: Right now I don't see a reason for an update of the spec at all. There is no new feature/functionality (unless I missed something)
If an update is needed for whatever reason, then yes 2.1 is enough
And I strongly encourage to avoid a major version change as this is a breaking change and makes it impossible (close to impossible) to provide an implementation support both versions. That's why we created a new package space when changing the spec to support Jakarta servlets. This allows to run javax servlets side-by-side with jakarta servlets.
The servlet API has breaking changes from 5 to 6 (which can mostly be neglected) but that's not OSGi's concern here.
I would also avoid tying a version to a servlet api version, meaning avoid to say that 2.0 can only be used with servlet 5 and 2.1 (if really required) can only be used with servlet 6.
Whiteboard 2.0 works happily with servlet API 6, no issue there. The Apache Felix Jetty12 Implementation passes all OSGi tests from R8.1 (and earlier): http service, http whiteboard, servlet whiteboard.

PR #613, opened 2023-08-03
Design for Type Safe Events 1.1

timothyjward:
This PR includes API, specification and test updates for the Type Safe Events 1.1 specification

### DIFF
--- a/cnf/ext/central.mvn
+++ b/cnf/ext/central.mvn
@@ -187,7 +187,7 @@ org.slf4j:slf4j-simple:1.7.7
 org.slf4j:log4j-over-slf4j:1.7.7
 
 org.apache.aries.component-dsl:org.apache.aries.component-dsl.component-dsl:1.2.2
-org.apache.aries.typedevent:org.apache.aries.typedevent.bus:0.0.1-RC1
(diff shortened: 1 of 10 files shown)


timothyjward:
TODO - we should discuss whether Record types are permitted as events

stbischof:
if wou want to handle Java 16 Records without relying on Java 16: (reflectiva Access) 
https://github.com/stbischof/osgi/commit/0818b5981ea37b8ff799dc2f6b9587e6e58b5515

timothyjward:
This PR has been updated with the changes requested by @jjh-aicas 

I have a couple of open questions:

* Should the return value of `getMaximumEventStorage` be an `int` rather than a `long`?
* Are we happy to use `Map.Entry` for the history retention policy or should we make a type just to hold the two values?
* Does the API (minimum and maximum retention) meet all the requirements, especially the return value from `configureHistoryStorage`?
* Are the topic Predicate/test methods useful?

stbischof:
for TCK we have to extend this.

https://github.com/osgi/osgi/tree/main/org.osgi.test.cases.typedevent

PR #633, opened 2023-11-15
Describe new  DS header that controls SCR behavior on lazy bundles

laeubi:
See discussion in
- https://github.com/osgi/osgi/discussions/590


laeubi:
Could the test-failure be related to my change?

stbischof:
i rerun the failing jobs

stbischof:
might be a good idea to ask @tjwatson on his thoughts about that. that we could have a more focused discussion on next spec call.

laeubi:
Still an issue

PR #657, opened 2024-01-15
[dependency] update felix.scr to 2.2.6

stbischof:


tjwatson:
This does not fail for me locally.  I see the tests failing in other PRs where the update to 2.2.6 was not made.  I don't think the update to 2.2.6 caused the test to fail during the build.

stbischof:
Correct. The test is failing also with 2.2.0-RC.

This was just my test to see if this may help.

tjwatson:
Do we know what 3 tests fail?  Where do we get the junit test results?

tjwatson:
There is a bug in the version of Java 17 being used that is causing an NPE to happen where an `IllegalArgumentException` use to be thrown from the ProxyBuilder.  On another version of Java 17 (semeru Java) I see this exception:

```java
java.lang.IllegalArgumentException: org.osgi.test.cases.component.service.BaseService referenced from a method is not visible from class loader
	at java.base/java.lang.reflect.Proxy$ProxyBuilder.ensureVisible(Proxy.java:883)
	at java.base/java.lang.reflect.Proxy$ProxyBuilder.validateProxyInterfaces(Proxy.java:721)
	at java.base/java.lang.reflect.Proxy$ProxyBuilder.<init>(Proxy.java:648)
	at java.base/java.lang.reflect.Proxy$ProxyBuilder.<init>(Proxy.java:656)
	at java.base/java.lang.reflect.Proxy.lambda$getProxyConstructor$0(Proxy.java:429)
	at java.base/jdk.internal.loader.AbstractClassLoaderValue$Memoizer.get(AbstractClassLoaderValue.java:329)
	at java.base/jdk.internal.loader.AbstractClassLoaderValue.computeIfAbsent(AbstractClassLoaderValue.java:205)
	at java.base/java.lang.reflect.Proxy.getProxyConstructor(Proxy.java:427)
	at java.base/java.lang.reflect.Proxy.newProxyInstance(Proxy.java:1037)
	at org.osgi.test.support.mock.MockFactory.newMock(MockFactory.java:96)
	at org.osgi.test.cases.component.junit.DeclarativeServicesControl.testComponentServiceObjectsBundle130(DeclarativeServicesControl.java:7463)
```

So we do see an `IllegalArgumentException` but it is not thrown from the expected call to `org.osgi.service.component.ComponentServiceObjects.ungetService(Object)` but instead to the `newMock` call.  I updated the `newMock` calls to use an anonymous inner class instead by using `new Object() {}` to the call to `newMock` to force it to be from a class loader that has visibility to the `BaseService` class.  This avoided the mock issue.  But now it exposed a real bug in SCR where the `ComponentServiceObjects.ungetService` implementation is not throwing the `IllegalArgumentException`. 

It appears we have never had a proper test for this assertion and Felix SCR has always had this bug.

tjwatson:
I got PR #666 going to fix the TCK.  Closing this PR.

PR #659, opened 2024-01-15
Design/587

laeubi:


org.osgi.test.cases.cm/.classpath
```diff
@@ -3,5 +3,6 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="aQute.bnd.classpath.container"/>
 	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
```
stbischof, line 6: not part of this spec pr

org.osgi.test.cases.jpa/.classpath
```diff
@@ -3,5 +3,6 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="aQute.bnd.classpath.container"/>
 	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
```
stbischof, line 6: rm this

stbischof (changes requested):

laeubi:
A preview of the proposed specification text can be found here: [osgi.cmpn-8.1.0.pdf](https://github.com/user-attachments/files/16163074/osgi.cmpn-8.1.0.pdf)

I also added TCKs for a simple webservice registration and the both Handlertypes now, to prove basic implementation of the spec here is an implementation of the spec that fulfills the current TCK

https://github.com/laeubi/xmla-services-workground/tree/osgi_spec_impl/ws/runtime/registrar

org.osgi.test.cases.jakartaws/src/org/osgi/test/cases/jakartaws/junit/SoapWihteboardTestCase.java
```diff
@@ -0,0 +1,109 @@
+package org.osgi.test.cases.jakartaws.junit;
```
stbischof, line 1: Please add a license header.

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/whiteboard/SoapWhiteboardConstants.java
```diff
@@ -0,0 +1,72 @@
...
+	 * Base prefix used in component property types
+	 */
+    public static final String SOAP_PREFIX = SOAP+".";
+	/**
+	 * Prefix used for properties of an endpoint implementor
+	 */
+	public static final String	SOAP_ENDPOINT_PREFIX		= SOAP_PREFIX
+			+ "endpoint.";
+
+	/**
```
stbischof, line 45: Please replace the TODO with a proper Documentation

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/dto/HandlerDTO.java
```diff
@@ -0,0 +1,31 @@
...
+import org.osgi.framework.dto.ServiceReferenceDTO;
+
+/**
+ * Represents a handler currently known to the webservice runtime
+ */
+public class HandlerDTO extends DTO {
+	/**
+	 * The service reference of the handler, is never {@code null}.
+	 */
+	public ServiceReferenceDTO serviceReference;
```
stbischof, line 30: maybe just use the  `serviceId` like in filter

https://github.com/osgi/osgi/blob/b32032ea3bb23de7d3b27622dc3865dfc803cef5/org.osgi.service.servlet/src/org/osgi/service/servlet/runtime/dto/FilterDTO.java#L87C1-L96C1

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/dto/RuntimeDTO.java
```diff
@@ -0,0 +1,46 @@
...
+	 * known, The returned array may be empty.
+	 */
+	public EndpointDTO[]			endpoints;
+
+	/**
+	 * Returns all handlers that are known and possible bound to an endpoint,
+	 * the returned array may be empty.
+	 */
+	public HandlerDTO[]			handlers;
+
```
stbischof, line 45: public  FailedHandlerDTO failedHandler

Like in ServletSpec
  https://github.com/osgi/osgi/blob/b32032ea3bb23de7d3b27622dc3865dfc803cef5/org.osgi.service.servlet/src/org/osgi/service/servlet/runtime/dto/DTOConstants.java#L62C1-L66C68

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/dto/RuntimeDTO.java
```diff
@@ -0,0 +1,46 @@
...
+public class RuntimeDTO extends DTO {
+
+	public ServiceReferenceDTO	serviceReference;
+
+	/**
+	 * Returns the representations of the Web Services endpoints currently
+	 * known, The returned array may be empty.
+	 */
+	public EndpointDTO[]			endpoints;
+
```
stbischof, line 39: add


public FailedEndpointDTO[]			failedEndpoints;

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/dto/HandlerDTO.java
laeubi: A service reference can easily transformed into a DTO and offer much more information (e.g. properties) as the requested error reporting capabilities will as well require to fetch service references it also seem to restrict the implementation to have the full information available here.

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/dto/HandlerDTO.java
```diff
@@ -0,0 +1,37 @@
...
+
+	/**
+	 * Returns the representations of the Web Services endpoints this handler is
+	 * currently bound to, the returned array may be empty.
+	 */
+	public EndpointDTO[]		boundEndpoints;
```
timothyjward, line 36: As described in https://github.com/osgi/osgi/pull/659/files#r1697146194 - this causes a cycle in the DTOs which is forbidden.

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/dto/RuntimeDTO.java
```diff
@@ -0,0 +1,62 @@
...
+	 * Returns the current service reference under that the runtime is
+	 * registered
+	 */
+	public ServiceReferenceDTO	serviceReference;
+
+	/**
+	 * Returns the representations of the Web Services endpoints currently
+	 * registered, The returned array may be empty.
+	 */
+	public EndpointDTO[]			endpoints;
```
timothyjward, line 42: If https://github.com/osgi/osgi/pull/659/files#r1697146194 is used then this would also have `BoundHandlerDTO[] boundHandlers`.

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/dto/RuntimeDTO.java
```diff
@@ -0,0 +1,62 @@
...
+	 * Returns the representations of the Web Services endpoints currently known
+	 * but failed to register, The returned array may be empty.
+	 */
+	public FailedEndpointDTO[]	failedEndpoints;
+
+	/**
+	 * Returns all handlers that are known and bound to an endpoint, the
+	 * returned array may be empty.
+	 */
+	public HandlerDTO[]			handlers;
```
timothyjward, line 54: As per https://github.com/osgi/osgi/pull/659/files#r1697146194 this would be `BoundHandlerDTO[]`

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/package-info.java
```diff
@@ -0,0 +1,50 @@
...
+ * {@code  Import-Package: org.osgi.service.jakartaws.runtime; version="[1.0,2.0)"}
+ * <p>
+ * Example import for providers implementing the API in this package:
+ * <p>
+ * {@code  Import-Package: org.osgi.service.jakartaws.runtime; version="[1.0,1.1)"}
+ * 
+ * @author $Id$
+ */
+
+@Version(SOAP_SPECIFICATION_VERSION + ".0")
```
timothyjward, line 38: The `.0` is unnecessary.

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/whiteboard/SoapWhiteboardConstants.java
```diff
@@ -0,0 +1,72 @@
...
+ * SPDX-License-Identifier: Apache-2.0 
+ *******************************************************************************/
+package org.osgi.service.jakartaws.whiteboard;
+
+/**
+ * Defines standard constants for the Jakarta Web Services Whiteboard services.
+ * 
+ * @author $Id$
+ */
+public class SoapWhiteboardConstants {
```
timothyjward, line 25: Is it only SOAP that is supported? Isn't the jakarta specification technically independent of SOAP, and allows XML/HTTP binding without SOAP?

laeubi: SOAP is an envelope format based on XML. Webservices defines the Exchange of SOAP Messages via a given transport (that can be HTTP or JMS or ...) the [specification website](https://projects.eclipse.org/projects/ee4j.jaxws) says:

> The Jakarta XML Web Services is a Java programming language API for creating web services, particularly SOAP services.

laeubi: If you are interested in the full details this gives a good overview:

https://jakarta.ee/specifications/xml-web-services/4.0/jakarta-xml-ws-spec-4.0.html#introduction

basically one could also use XML-RPC its been a while since I've seen any (working) implementation for modern java, the official RI seems this one and jax-ws only reference that as an alternative binding https://github.com/eclipse-ee4j/jax-rpc-ri

maybe because most of the time SOAP (that supports Message and RPC style) is mentioned as the successor of XML-RPC. At least this specification stems from an implementation that based on SOAP and therefore has this naming.

laeubi: I have now renamed all SOAP > Webservice

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/package-info.java
laeubi: I have taken this pattern from the REST implementation.
My guess is / was that the idea is that the spec version always stays the same (1.0) and the last part is possibly incremented (e.g. SOAP_SPECIFICATION_VERSION + ".1") in case some API enhancements are performed that require an update to the package but not to the specification version.

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/runtime/dto/EndpointDTO.java
```diff
@@ -0,0 +1,45 @@
...
+	 * The full resolved address this endpoint is published to This value is
+	 * never {@code null}.
+	 */
+	public String				address;
+
+	/**
+	 * Returns all handlers that are bound to this endpoint, the returned array
+	 * may be empty.
+	 */
+	public HandlerDTO[]			handlers;
```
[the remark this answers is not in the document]
laeubi: I removed the link form `Handler` > `Endpoint` as the information is already implicitly included:

If I want to know what `Endpoint`s a given `Handler` is bound I can simply iterate over all endpoints and check their `Handler` array if it contains the handler of interest e.g. trivially by using the service.id as a key and a list of endpoint

laeubi:
I updated the spec to reflect the recent changes [osgi.cmpn-8.1.0.pdf](https://github.com/user-attachments/files/16440939/osgi.cmpn-8.1.0.pdf)

org.osgi.service.jakartaws/src/org/osgi/service/jakartaws/whiteboard/WebserviceWhiteboardConstants.java
```diff
@@ -0,0 +1,74 @@
...
+public class WebserviceWhiteboardConstants {
+
+    private WebserviceWhiteboardConstants() {
+        // non-instantiable
+    }
+
+	/**
+	 * Base namespace for the Webservice Whiteboard specification
+	 */
+	public static final String	WEBSERVICE							= "osgi.ws";
```
stbischof, line 34: osgi.ws
osgi.jakartaws
osgi.jakartaxmlws
osgi.jakarta.ws
oagi.jakarta.xml.ws

@timothyjward 
@laeubi is this in line with what we jave in the other jakarta specs?

laeubi: https://github.com/osgi/osgi/blob/b32032ea3bb23de7d3b27622dc3865dfc803cef5/org.osgi.service.jakartars/src/org/osgi/service/jakartars/whiteboard/JakartarsWhiteboardConstants.java#L220

stbischof: osgi.jakarta.xml.ws

laeubi:
I updated the naming discussed now and updated the specification [osgi.cmpn-8.1.0.pdf](https://github.com/user-attachments/files/16614150/osgi.cmpn-8.1.0.pdf)

laeubi:
I have now moved my TCK and reference implementation to this repository:

- https://github.com/laeubi/osgi-xml-webservice

this should make it able for the build to succeed here.

laeubi:
@stbischof @timothyjward I think I need some help here it looks like my new added `jakarta.xml.soap-api` (version 3) now conflicts somehow with CDI (version 1.4) both are in the central index and I would have assumed bnd to sort that out but it seems not the case.

laeubi:
- [ ] Shorten examples by using multi-catch
- [ ] Make clear that registering a (invalid) handler will result in an error
- [ ] Add a property type that enables the pickup of `Handler` interfaces by the whiteboard / probably combine with the filter expression type
- [ ] jax-ws-ap --> jakarta xml webservices API
- [ ] runtime API doc is missing

org.osgi.service.webservice/src/org/osgi/service/webservice/whiteboard/propertytypes/WhiteboardHandler.java
```diff
@@ -0,0 +1,63 @@
...
+ * Webservices Whiteboard Runtime
+ */
+@Documented
+@Retention(RetentionPolicy.CLASS)
+@Target({
+		ElementType.TYPE
+})
+@RequireWebserviceWhiteboard
+@ComponentPropertyType
+public @interface WhiteboardHandler {
```
stbischof, line 41:
```suggestion
public @interface XmlWsHandler {
```

stbischof: also in the docbook

stbischof (changes requested):

laeubi:
I have published the implementation here:

- https://github.com/eclipse-osgi-technology/jakarta-webservices/pull/1

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -0,0 +1,394 @@
...
+         </section>
+         <section>
+            <title>The Endpoint Publisher</title>
+            <para>
+            The responsibility of the <emphasis>Endpoint Publisher</emphasis> is to use the <emphasis>Endpoint</emphasis> and publish them with a given transport.
+            An implementation might use different techniques to discover a suitable way and support for different protocols but at least the publishing using the default enpoint publish must be supported, in addition to that implementations are encouraged
+            to also support publishing to the Http Whiteboard Service using property <emphasis>osgi.jakarta.xml.ws.endpoint.http.contextpath</emphasis> as described in <emphasis>Registering Endpoint Implementor</emphasis>
+            </para>
+        </section>
+        <section>
```
timothyjward, line 242: We are missing a section describing the `WebserviceServiceRuntime` service. I assume that this will manage a `changecount` property like the other whiteboards? We also need descriptions of what the DTO fields represent.

laeubi:
> There are also issues with the implementation, which currently does not work due to incompatibilities between metro and JAXB.

I don't understand this, as I also wrote some TCKs for the implementation and also @stbischof has used the impl already im quite confident it works, but be aware that you likely want the `org.glassfish.hk2.osgiresourcelocator` in your setup to let all the jakarta stuff find its implementations. (see also https://github.com/osgi/osgi/issues/685 where I proposed a more general way to make OSGi integrate better wit J2EE)

stbischof:
Could you link the latest test and setup you made Christoph?

timothyjward:
> > There are also issues with the implementation, which currently does not work due to incompatibilities between metro and JAXB.
> 
> I don't understand this, as I also wrote some TCKs for the implementation and also @stbischof has used the impl already im quite confident it works, but be aware that you likely want the `org.glassfish.hk2.osgiresourcelocator` in your setup to let all the jakarta stuff find its implementations. (see also #685 where I proposed a more general way to make OSGi integrate better wit J2EE)

My test registers an "Echo" service identical to the one in the specification:

- It fails to register the endpoint due to "missing return" on a generated class. This turns out to be because the test bundle doesn't import any JAXB packages. If I add `DynamicImport-Package: jakarta.xml.bind.*` then it progresses further
- It fails to register the endpoint due to Metro being unable to set up a JAXBContext. This appears to be due to JAXB trying to load a class from Metro without using the correct class loader.

timothyjward:
> Could you link the latest test and setup you made Christoph?

I'm pushing my work in progress to https://github.com/timothyjward/osgi/tree/tw/jakartaws-ct

laeubi:
> Could you link the latest test and setup you made Christoph?

Yes I will add some more code to the implementation, I just wanted to mention that metro is/was not used at all, this is actually just some delegation of the jakarta stuff that is not really usefull here, we want to simply use the reference implementations.

timothyjward:
> Yes I will add some more code to the implementation, I just wanted to mention that metro is/was not used at all, this is actually just some delegation of the jakarta stuff that is not really usefull here, we want to simply use the reference implementations.

Normally you will need to pick an implementation, or put a *lot* of effort into supporting different implementations. The servlet whiteboard uses Jetty, the Jakarta REST whiteboard uses Jersey, neither of them are pluggable. This allows them to fix or work around issues to that users of the whiteboard don't have to. 

Aries JPA tries to be pluggable, and it's very hard to maintain as a result.

laeubi:
@timothyjward I'm not sure I understand the reference implementation already **is** an implementation and it works, of course one can decide to use other implementations as well but most of them are not very integrated with OSGi or have very large dependency chains and require special configuration. Aynways maybe things become clearer once I add the integration tests to the implementation repo.

timothyjward:
> @timothyjward I'm not sure I understand the reference implementation already **is** an implementation and it works, of course one can decide to use other implementations as well but most of them are not very integrated with OSGi or have very large dependency chains and require special configuration. Aynways maybe things become clearer once I add the integration tests to the implementation repo.

Sorry if I wasn't clear - normally when implementing a whiteboard specification such as this one, the OSGi whiteboard implementation would have to target a specific implementation of the Jakarta specification. For example, the Felix servlet whiteboard implementation uses Jetty directly, and cannot use TomCat or Grizzly.

My expectation is that an implementation of the XML Web Services whiteboard will have to make a similar decision, and be tied to a specific implementation of `jakarta.xml.ws`. Otherwise the overhead of working around bugs becomes an exercise for the user of the whiteboard, which can be very painful.

laeubi:
> Could you link the latest test and setup you made Christoph?

The runnable test are now here:
- https://github.com/eclipse-osgi-technology/jakarta-webservices/pull/7

timothyjward:
The current draft has a section titled _"The_ Endpoint Publisher"_ which describes a pluggable publisher. I can't find any interface which matches this in the specification, nor anything testable. 

The specification must either:
* Fully specify the integration point
or:
* Remove this section and leave the publication as an implementation detail

I would be tempted to remove this section, and simply to say that the Endpoint Registrar is responsible for ensuring that the endpoint is published in some way. We can always expand to a publisher whiteboard in a version 1.1 if there is a demand for it.

osgi.specs/docbook/160/service.jakartaws.xml
laeubi: I added changecount, what the DTO fields represent should be described in the javadoc, is it really useful to replicate that information?

laeubi:
I have rewritten the section about  Endpoint Publisher now

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -0,0 +1,394 @@
...
+                    then the Endpoint is considered failed.
+                    The error code in this case is <code>203</code>.
+                    </para>
+                  </listitem>
+                </itemizedlist>
+            </section>
+          </section>
+                
+
+
```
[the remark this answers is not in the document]
laeubi: I have now added the Capabilities section.

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -355,7 +355,55 @@ public class SOAPLoggingHandler implements SOAPHandler&lt;SOAPMessageContext&gt;
...
+    <section>
+      <title>osgi.implementation Capability</title>
+
+      <para>The Jakarta™ XML Web Services Whiteboard Implementation bundle must provide the <link
+      linkend="service.namespaces-osgi.implementation.namespace"><code>osgi.implementation</code></link>
+      capability with the name <xref
+      linkend="org.osgi.service.webservice.whiteboard.WebserviceWhiteboardConstants.WEBSERVICE_IMPLEMENTATION"
+      xrefstyle="hyperlink"/>. This capability can be used by provisioning
+      tools and during resolution to ensure that a Typed Event implementation
```
timothyjward, line 369: This should be Web Services Whiteboard implementation I think?

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -355,7 +355,55 @@ public class SOAPLoggingHandler implements SOAPHandler&lt;SOAPMessageContext&gt;
...
+      is present. The capability must also declare a uses constraint for the
+      <code>org.osgi.service.typedevent</code> package and provide the version
```
timothyjward, line 371: I think that this should be the packages `org.osgi.service.webservice.whiteboard`, `jakarta.xml.ws` and `jakarta.xml.ws.handler`. We possibly also need `jakarta.jws` and `jakarta.jws.soap` but I'm not so sure of those

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -355,7 +355,55 @@ public class SOAPLoggingHandler implements SOAPHandler&lt;SOAPMessageContext&gt;
...
+      of this specification:</para>
+
+      <programlisting>Provide-Capability: osgi.implementation;
+ osgi.implementation=osgi.webservice";
+ uses:="org.osgi.service.webservice.runtime,org.osgi.service.webservice.runtime.dto";
```
timothyjward, line 376: I think these are the wrong packages. See above

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -355,7 +355,55 @@ public class SOAPLoggingHandler implements SOAPHandler&lt;SOAPMessageContext&gt;
...
+      linkend="service.namespaces-osgi.implementation.namespace"/>.</para>
+    </section>
+
+    <section>
+      <title>osgi.service Capability</title>
+
+      <para>The bundle providing the Jakarta™ XML Web Services Whiteboard service must provide
+      capabilities in the <link
+      linkend="service.namespaces-osgi.service.namespace"><code>osgi.service</code></link>
+      namespace representing the services it is required to register. This
```
timothyjward, line 395: I think that this is just the Runtime service, but we should be explicit.

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -239,6 +244,51 @@ public class SOAPLoggingHandler implements SOAPHandler&lt;SOAPMessageContext&gt;
...
+              of this property will increase.</para> <para>This allows
+              interested parties to be notified of changes to the DTOs by
+              observing Service Events of type <code>MODIFIED</code> for the
+              <xref linkend="org.osgi.service.webservice.runtime.WebserviceServiceRuntime"
+              xrefstyle="hyperlink"/> service. See <code>org.osgi.framework.Constants.SERVICE_CHANGECOUNT</code> in <xref linkend="intro.core.release" xrefstyle="template:%t"/>.</para></entry>
+            </row>
+          </tbody>
+        </tgroup>
+      </table>
+        </section>
```
timothyjward, line 279: This section should have a subsection describing the method for obtaining a DTO, and its rules:

- Is the DTO live or a snapshot (hopefully a snapshot!)
- What is stored in the returned DTO (endpoints, handlers, failures)

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -0,0 +1,480 @@
...
+                    In all cases an additional error message might be used to further describe the nature of the failure and give users of the DTO a hint what's going on.
+                    </para>
+                  </listitem>
+                  <listitem>
+                    <para><emphasis>Failure to obtain a service instance</emphasis> - In
+                    the case where a published service is unable to be obtained by the
+                    Jakarta Webservice whiteboard then the service is deny
+                    listed by the container. A failure DTO is made available from the
+                    WebserviceServiceRuntime representing the deny listed service
+                    object. The error code in this case is <code>1</code>.</para>
```
timothyjward, line 308: There should be constants for the defined error codes. See https://github.com/osgi/osgi/blob/main/org.osgi.service.servlet/src/org/osgi/service/servlet/runtime/dto/DTOConstants.java as an example

laeubi: See https://github.com/osgi/osgi/pull/659/files#diff-3156a92db7e047bef4da747c15e1ffb4a7a9f0754cd1eea27b2d815601e84362R36

timothyjward: It's good that they're there already. Please link them in the spec text so that they're easier to find 😃

osgi.specs/docbook/160/service.jakartaws.xml
timothyjward: > is it really useful to replicate that information?

There are parts not easily visible in the javadoc, for example:
* Is the DTO is a snapshot or a live object
* Is the DTO a copy, or are instances shared?
* Why might I use the RuntimeDTO, ideally including an example.

laeubi: [DTO spec](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.dto.html) says a DTO must be

> - Snapshot - Does not track changes to represented runtime object.
> - Not Thread Safe - External synchronization is needed.

> Why might I use the `RuntimeDTO`, ideally including an example.

If you have no use for it, don't use it ;-)

In any case if things are missing I better would like to enhance the javadoc, because that is what users most likely will read and what is most visible in your IDE.

timothyjward: > In any case if things are missing I better would like to enhance the javadoc, because that is what users most likely will read and what is most visible in your IDE.

OSGi has a long history of producing good quality specification documents that are readable and understandable. They cover the necessary cases for implementors to reliably produce a compatible implementation, and the information users need to understand why things behave as they do. Good JavaDoc is important, and helpful in many cases, but the specification text is the "source of truth" and the place we refer back to whenever there is a question from a user or implementer.

> If you have no use for it, don't use it ;-)

If I'm new to the specification I might not know that I do have a use for it. Examples are a good thing and should be used regularly.

laeubi: > OSGi has a long history of producing good quality specification documents that are readable and understandable.

I don't see that javadoc is "less understandable" also the javadoc is part of / included in the specification. So I don't see we lose anything (beside the risk of having two equal text blocks getting out of sync).

timothyjward: > I don't see that javadoc is "less understandable" also the javadoc is part of / included in the specification.

It's included but separate. You lose context by it being in a different place, and JavaDoc has to be written in a code-centric fashion (because it's attached to the code) whereas a specification is written in a solution-centric fashion.

timothyjward:
> > What if they have different namespaces or signatures but the same base path?
> 
> I don't understand this part, what do you mean by "namespace" or "signature", the namespace in a web-service is the URI so only one item can ever have the same URI (in contrast to REST) as they all use the POST verb.

The XML in the SOAP message is namespaced, you could (at least theoretically) use the namespace to pick the correct endpoint to call if the namespaces were different. I know that the namespace is used for identity, because if you include the wrong one then invocation fails, but my knowledge isn't deep enough to go further than that.

> 
> > If one wins then can I control which one?
> 
> First registered wins.

So the service ranking is ignored? Whatever the rule is it needs to be described in the specification so that people know what behaviour to expect.

laeubi:
> The XML in the SOAP message is namespaced, you could (at least theoretically) use the namespace to pick the correct endpoint to call if the namespaces were different.

I'm not sure how this is relevant for the specification. Actually there are only two cases:

1. You can publish the endpoint to the given address
2. The publishing fails

Why it fails can have numerous reasons that are not possible, e.g. because an transport supports multiple endpoints at the same address, the protocol is not supported or whatever (e.g. missing annotations)

laeubi:
> What happens if I register two endpoints for the same address?
> 
>     * What if they have different namespaces or signatures but the same base path?
> 
>     * If one wins then can I control which one?

I have now added the following to the spec:


>>      There is no particular order in which Endpoint implementors are published, it could be that they are registered as soon as the service is visible in the OSGi service registry
>>      or even delayed in the background. If any particular ordering of publishing is desired this must be synchronized externally, e.g by observing the Runtime Service and waiting for a prerequisite
>>      endpoint to show up. Due to the dynamic nature of OSGi, most system should be design in a way that the exact point in time of publishing the endpoint is not required for the system to function correctly.

This should make clear that any control is not possible at all and is generally an error if the order matters, If the underlying transport allows publishing of multiple endpoint (what is quite unlikely for the common ones) it would be supported. If not then any of those will fail in possibly random order,

timothyjward:
> I'm not sure how this is relevant for the specification.

Well a common case in OSGi is that two services are competing to be registered at the same location. There are explicit rules for what happens in this case in the Servlet and Jakarta REST whiteboards. In both of these specifications they make it clear that whiteboard services are registered in *service rank order*, meaning that you can "choose" which service wins by changing its ranking. 

Remaining silent in these cases is less helpful as it can lead to different implementations making different decisions about what should happen, reducing the portability of applications between implementations. I'm only asking for similar precision in this specification as exists in similar existing whiteboards.

timothyjward:
It looks like our comments crossed. What you've suggested is a perfectly reasonable approach, albeit inconsistent with other whiteboards. We can raise it for feedback from others at the next meeting.

laeubi:
> * Is it possible to control the order in which Handler services are applied? If so how?

I have now added the following section to the spec:

>>  Handlers will be added to the Endpoint in service ranking order, that means that Handlers with a higher service ranking will be called before those with a lower service ranking.
 If two handlers have the same ranking they are sorted according to their service id, having handlers registered later being called later in the handler chain.

osgi.specs/docbook/160/service.jakartaws.xml
laeubi: Done

osgi.specs/docbook/160/service.jakartaws.xml
laeubi: Done, I updated the initial description with the current version of the spec.

timothyjward:
As requested in the call - the name of the branch where the CT is running is https://github.com/timothyjward/osgi/tree/tw/jakartaws-ct

laeubi:
@timothyjward can we have a checklist here what needs adjustments until we are done?

timothyjward:
> @timothyjward can we have a checklist here what needs adjustments until we are done?

These are the failing tests:

- [ ] eclipse-osgi-technology/jakarta-webservices/issues/18
- [ ] eclipse-osgi-technology/jakarta-webservices/issues/17
- [ ] eclipse-osgi-technology/jakarta-webservices/issues/9

I *think* that the TCK is now complete and that I should merge it somewhere so that we start getting test runs in CI. What is the best place to do that? Is it `main` (which would involve this feature being merged first)? Is it `design/587`, and if so do I need to be careful about history? Is it somewhere else (if so what should I use as my fork point)?

stbischof:
new snapshots of the impl are deployed to

Sonatype Central Snapshot
```
   <repository>
      <id>central-snapshots</id>
      <name>Central Snapshot</name>
      <url>https://central.sonatype.com/repository/maven-snapshots/</url>
      <releases>
        <enabled>false</enabled>
      </releases>
      <snapshots>
        <enabled>true</enabled>
      </snapshots>
    </repository>
 ```

Link to push
https://github.com/eclipse-osgi-technology/jakarta-webservices/actions/runs/16197999204/job/45729433488#step:5:1373

laeubi:
I rebased and force pushed my branch as well

PR #661, opened 2024-01-17
Ensure delegate object class loader has visibility to service class

tjwatson:
This PR will fail the component TCK until we get a Felix SCR release with the fix for https://issues.apache.org/jira/browse/FELIX-6682 included

### DIFF
--- a/org.osgi.test.cases.component/src/org/osgi/test/cases/component/junit/DeclarativeServicesControl.java
+++ b/org.osgi.test.cases.component/src/org/osgi/test/cases/component/junit/DeclarativeServicesControl.java
@@ -7346,7 +7346,7 @@ public class DeclarativeServicesControl extends DefaultTestBundleControl
 				assertEquals("wrong number of services", 1, baseReg.getReference().getUsingBundles().length);
 
 				try {
-					cso.ungetService(MockFactory.newMock(BaseService.class, new Object()));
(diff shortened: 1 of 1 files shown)


laeubi:
@tjwatson can you explain the difference between `new Object()` and `new Object() {}` here?
Is it to have a (private) implementation object?

org.osgi.test.cases.component/src/org/osgi/test/cases/component/junit/DeclarativeServicesControl.java
```diff
@@ -7460,7 +7460,7 @@ public void testComponentServiceObjectsBundle130() throws Exception {
 				}
 				assertEquals("wrong number of services", 1, counter.get());
 				try {
-					cso.ungetService(MockFactory.newMock(BaseService.class, new Object()));
+					cso.ungetService(MockFactory.newMock(BaseService.class, new Object() {}));
```
laeubi, line 7463: Maybe this can be replaced with `assertThrows`?

tjwatson:
> @tjwatson can you explain the difference between `new Object()` and `new Object() {}` here?
> Is it to have a (private) implementation object?

The class loader used for the implementation class.  That is, the system (null) class loader vs. the bundle class loader of the test bundle.  The null system class loader which loads the `Object` class does not have access to the service interface while the test bundle class loader does.  And the null class loader causes an NPE on the version of Java 17 being used in the build.

Issue #669, opened 2024-01-26
Converter implemetation bug due to aparent typo.


folarte:
I'm not familiar enough with git to make a pull request, but I think this can be easily solved.

In lines 73-74 of method `InternalConverting applyModifiers(InternalConverting converter)` on class `org.osgi.util.converter.FunctioningImpl` 
```
		if (targetAsDTO)
			converter.targetAsBean();
 ```
it seems `targetAsBean()`is called instead of `targetAsDTO()`.

I have looked at apache felix sources too and bug seems to be there too. I noticed it when a conversion
```
       return Converters.standardConverter().convert(properties)
                    .targetAsDTO()
                    .to(SqsSourceParams.class);
```
Worked flawlessly but 
```
    Function<Object, SqsSourceParams> cvt
            = Converters.standardConverter().function()
                    .targetAsDTO()
                    .to(SqsSourceParams.class);
     return cvt.apply(properties);
```
failed, and tracing it revealed my class was not being treated as a DTO.

folarte:
Sorry for the delay. Totally lost in github. I have managed to find the PR, looked at the commit and submitted an approval to it. I do not know if this is OK. I also do not know if I should close this.

PR #681, opened 2024-02-18
[converter] fix conv. to Function with DTO

stbischof:
Fix :https://github.com/osgi/osgi/issues/669


folarte (approved):
One line modification is exactly what I reported, test case seems to prove it. I do not have enough github knowledge to say more and I have not been able to set up a full build chain.
The test is extremely nice, seeing the whole file it seems original author deemed the asXXXX functions or other things to trivial to test and the  typo slipped.
I do not know if I should approve, but as the check box is enabled I will set it.

folarte (approved):
Supression is exactly what I spotted and commented, seems totally ok to me.

stbischof:
thank you @folarte

will wait for @timothyjward  approve and then merge.
maybe @timothyjward has an ides on next release of thios micro bugfixing  version

Issue #718, opened 2024-04-27
[tck][cm] CMCoordinationTestCase#testCoordinatedConfigurationOnBeforeRegisteredManagedService seems to contain to strict assertions


laeubi:
Currently Equinox fails this Coordinator TCK test-case:

https://github.com/osgi/osgi/blob/11d440e336ac2c22bade54105ed4ef2c9aab79d8/org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java#L384-L441

the reason is that here is is asserted that *exactly one* event is triggered:
https://github.com/osgi/osgi/blob/11d440e336ac2c22bade54105ed4ef2c9aab79d8/org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java#L439

because the test-case assumes that the block that is executed under the coordinator results in no event being generated as it creates a configuration and then afterwards delete it.

But [the specification says](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.cm.html#service.cm-coordinatorsupport) (even though its quite vague):

> If configurations are created, updated or deleted and an implicit coordination exists, the Configuration Admin service must delay notifications until the coordination terminates. However the configuration changes must be persisted immediately. Updating a Managed Service or Managed Service Factory and informing asynchronous listeners is delayed until the coordination terminates

What now happens with equinox that the `updated` method is called twice, once outside the coordination and one after the coordination terminates, because actually something *was* happening (configuration got deleted) and the "delayed notification at termination" is exactly this.

Given that a `ManagedService` must be prepared to be called multiple times with `null` configuration (e.g. when the ConfigurationAdmin is restarted) and the ConfigAdmin should be [stateless in regards to what it has offered to the service](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.cm.html#service.cm-updating.configuration), I think this is also a valid behavior.

I therefore think the assertion should be that the number of event is not larger than 2 and each value is `false`.

laeubi:
Still an issue

PR #719, opened 2024-04-27
More than one event should be allowed in case of coordinated delete

laeubi:
Fix https://github.com/osgi/osgi/issues/718


stbischof:
@laeubi please update also the spec.

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
```diff
@@ -436,8 +436,11 @@ public void updated(Dictionary<String, ? > properties)
 
 		// wait and verify listener
 		sleep();
-		assertEquals(1, events.size());
-		assertFalse(events.get(0));
+		assertTrue(events.size() <= 2);
```
chrisrueger, line 439: Why not **exactly** 2?

just for the record quoting the output of the [failed testcase](https://github.com/eclipse-equinox/equinox/pull/559/checks?check_run_id=24695693302):

```
expected:<1> but was:<2>
junit.framework.AssertionFailedError: expected:<1> but was:<2>
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.failNotEquals(Assert.java:329)
	at junit.framework.Assert.assertEquals(Assert.java:78)
	at junit.framework.Assert.assertEquals(Assert.java:234)
	at junit.framework.Assert.assertEquals(Assert.java:241)
	at junit.framework.TestCase.assertEquals(TestCase.java:384)
	at org.osgi.test.cases.cm.junit.CMCoordinationTestCase.testCoordinatedConfigurationOnBeforeRegisteredManagedService(CMCoordinationTestCase.java:439)
```

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
```diff
@@ -436,8 +436,11 @@ public void updated(Dictionary<String, ? > properties)
...
+		// all events must be 'null' updates
+		for (Boolean b : events) {
```
chrisrueger, line 441: would it make sense to to additionally add an addition `assertEquals(1, events.size());` right after `registerService(..)` but outside coordination, to make it clearer that updated() should have been called?

image

chrisrueger (commented):
left some comments.

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
stbischof: Then felix command will fail.

laeubi: As the spec now mandates this I updated the testcase accordingly.

org.osgi.test.cases.cm/src/org/osgi/test/cases/cm/junit/CMCoordinationTestCase.java
laeubi: I have adjusted the testcase to be more expressive now.

laeubi:
I tried to address the comments and adjusted the spec here is the generated specification: [osgi.cmpn-8.1.0.pdf](https://github.com/osgi/osgi/files/15258031/osgi.cmpn-8.1.0.pdf)

osgi.specs/docbook/104/service.cm.xml
```diff
@@ -2464,6 +2464,30 @@ ConfigurationPermission[?application,<xref
     fails or terminates regularly. Registered synchronous listeners will be
     informed immediately when the change happens regardless of a
     coordination.</para>
+    <para>The intend of this integration is that multiple events are collapsed
+    into one as updating a Managed Service or Managed Service Factory might result
+    in services being updated or unregistered what can trigger other service changes
+    what might be a costly operation and therefore result in an unwanted intermediate
+    states until the system has setled again. Collapsing here means, that multiple
```
chrisrueger, line 2471: Spelling: 

> an unwanted intermediate states

Correction:  an unwanted intermediate state

(s removed at the end)

chrisrueger (commented):
- I like the clarifications in the testcase. 
- I suggested a small typo correction

I just comment (instead of approve) because I don't feel I am qualified enough to judge this, since I was late in the game. But it sounds reasonable and makes sense to me.

osgi.specs/docbook/104/service.cm.xml
chrisrueger: thanks good to know.

laeubi:
Still an issue ...

stbischof:
lets bring this up next spec call. 
will you also do the impementation ?

laeubi:
This is how it is implemented in Equinox ATM see:
- https://github.com/eclipse-equinox/equinox/issues/190

laeubi:
@stbischof @tjwatson I think we have already agreed on this in the spec call can we proceed here?

osgi.specs/docbook/104/service.cm.xml
```diff
@@ -2464,6 +2464,30 @@ ConfigurationPermission[?application,<xref
     fails or terminates regularly. Registered synchronous listeners will be
     informed immediately when the change happens regardless of a
     coordination.</para>
+    <para>The intend of this integration is that multiple events are collapsed
```
timothyjward:
```suggestion
    <para>The intent of this integration is that multiple events are collapsed
```

osgi.specs/docbook/104/service.cm.xml
```diff
@@ -2464,6 +2464,30 @@ ConfigurationPermission[?application,<xref
...
+    into one as updating a Managed Service or Managed Service Factory might result
+    in services being updated or unregistered what can trigger other service changes
```
timothyjward:
```suggestion
    in services being updated or unregistered which could then trigger additional service changes.
```

laeubi:
What needs to be done to get this thing finished?

HannesWell:
> What needs to be done to get this thing finished?

Yes, it would be great if this could be finished. The failure in Equinox is now happening since three years.

PR #744, opened 2024-09-10
Jakarta Whiteboard WebSocket Specification

laeubi:
I now have poked around the spec, checked the API and some reference implementaion for Jakarta Websockets. For the Sake of an OSGi Whiteboard Service it seems feasable to support only Annotated Endpoints as described in the spec as "annotated endpoint"

To support self-publishing (referenced in the spec as programmatic endpoint) as well, then I can add methods to the Runtime service to obtain necessary interface, I just wondering if this is a case we want to support at all.
One problem is that the specification mandates, that after the initial configuration phase no more server endpoints have to be registered anymore and that there is only one container per VM ... both things can obviously not promised in OSGi.

So given that, one can specify that OSGi Jakarta Websockets is **reusing** the spec and can handle endpoints annotated **according to the spec**, but such an implementation **can not be compliant** to the spec in its current form.

Another pitfall is, that one *can* support clients in a Whiteboard fashion, but it seems a bit odd to handle practically one has to register one service instance per client connection and somehow tie the connection state to that registration.

Next quirk is, that only **one** listener per type is supported on an endpoint, so having the whiteboard collecting listeners and there are possibly different variants of them so installing proxies seem not feasible at all as there is a high risk of clashes.

## Current Implementation:

An Implementation of the Spec in its current state is here:

https://github.com/laeubi/osgi-websockets

## Current Spec as PDF

Can be found here [osgi.cmpn-8.1.0.pdf](https://github.com/user-attachments/files/19068746/osgi.cmpn-8.1.0.pdf)


.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
+# Jakarta WebSocket Whiteboard
+
+This documents looks into the requirements for a Jakarta WebSocket whiteboard specification.
+
+## Introduction
+
+The OSGi Compendium R6 introduced the HTTP Whiteboard specification, upgraded in R8.1 to the Whiteboard specification for Jakarta Servlet.
+It provides a light and convenient way of using servlets in a an OSGi environment through the use of the Whiteboard Pattern.
+
+WebSocket is communication protocol designed to work over the HTTP protocol.
```
laeubi, line 10: This is wrong. see for example: https://en.wikipedia.org/wiki/WebSocket

> WebSocket is a computer [communications protocol](https://en.wikipedia.org/wiki/Communications_protocol), providing a [simultaneous two-way](https://en.wikipedia.org/wiki/Full-duplex) communication channel over a single [Transmission Control Protocol](https://en.wikipedia.org/wiki/Transmission_Control_Protocol) (TCP) connection.

and further down

> WebSocket is distinct from [HTTP](https://en.wikipedia.org/wiki/HTTP) used to serve most webpages. Although they are different, [RFC](https://en.wikipedia.org/wiki/RFC_(identifier)) [6455](https://datatracker.ietf.org/doc/html/rfc6455) states that WebSocket "is designed to work over HTTP ports 443 and 80 as well as to support HTTP proxies and intermediaries", thus making it compatible with HTTP. To achieve compatibility, the WebSocket [handshake](https://en.wikipedia.org/wiki/Handshake_(computing)) uses the [HTTP Upgrade header](https://en.wikipedia.org/wiki/HTTP/1.1_Upgrade_header)[[3]](https://en.wikipedia.org/wiki/WebSocket#cite_note-3) to change from the HTTP protocol to the WebSocket protocol.

So HTTP and Websocket are distinct protocols and HTTP supports switching of protocols (not only Websocket!).

The protocoll for websockets is `ws://` (or `wss://` for ssl), while HTTP uses `http://` and `https://`

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+It enables bidirectional exchange of messages between the web client and the server.
+Like an HTTP servlet, a WebSocket handler is bound to a URL.
+
+The conversion from an HTTP connection to a WebSocket stream is done by sending a `GET` HTTP request with a set of headers requesting a connection upgrade.
+The server will respond with a *101 Switching protocols* response.
+
+The WebSocket protocol is defined by [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455) and a Java specification is defined in Jakarta EE.
+This document considers the specification will rely on [Jakarta WebSocket 2.1](https://jakarta.ee/specifications/websocket/2.1/), from Jakarta EE 10.
```
laeubi, line 18:
```suggestion
This document considers the specification will rely on [Jakarta WebSocket 2.2](https://jakarta.ee/specifications/websocket/2.2/), from Jakarta EE 10.
```

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+
+HTTP server implementations can provide support for WebSocket, but their configuration is complex in the context of the OSGi Whiteboard specification for Jakarta Servlet.
```
laeubi, line 20: Not only Webservers, the [jakarata Refernece implementation ](https://github.com/eclipse-ee4j/tyrus) can work in a standalone mode as well.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+
+A Jakarta Websocket Whiteboard would ease the deployment of WebSocket handlers in an OSGi HTTP environment.
```
laeubi, line 22: As mentioned above, Websockets is not tied to HTTP, we might mention that integration with the HTTP Whiteboard is recommended but I don't think it is a requirement from a specification point of view.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+* *WebSocket message* - Data or control message transferred over the WebSocket. A message can be transmitted via one or multiple frames. A data message can be text or binary. A control message can be a ping, pong or close message.
+* *WebSocket Container* - The server-side, implementation-specific, container of server WebSocket Endpoints.
+* *WebSocket Session* - Description of a single client-server WebSocket connection.
+* *WebSocket Endpoint* - Handler of a WebSocket session lifecycle: open, close, error. The endpoint registers frame listeners when the session is opened. Can be used either on client or server side.
+* *WebSocket (Message) Handler* - Handler notified of WebSocket message reception. Can be used either on client or server side.
+
+## Problem Description
+
+Registering a WebSocket handler can currently be done by using a custom configuration of an HTTP servlet.
+The problem is that such configuration depends on the underlying HTTP server implementation.
```
laeubi, line 38: See previous comments, HTTP is not mandatory.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+
+For example, the Eclipse Jetty server requires registering components before initializing an HTTP servlet to allow it to handle the WebSocket protocol.
+This requires to write code specific to Jetty in an OSGi HTTP whiteboard environment.
+There is also an issue between the Jetty WebSocket server components and the servlet context given to it by the Apache Felix HTTP Jetty implementation.
```
laeubi, line 42: Jakarta Websockets include specific methods to interface with Sevlet API so these can be implemented in a container agnostic way.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+
+As a result, the registration of WebSocket in the current state of the specification goes against the loose coupling concept usually provided by OSGi.
+
+What is needed is a WebSocket whiteboard specification moving specific configuration to the HTTP/WebSocket whiteboard implementation.
```
laeubi, line 46: The HTTP Whiteboard should not be affected directly by this specification.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+Reusing parts, if not all, of the Jakarta WebSocket specification will ease the development of WebSocket handlers and avoid redefining existing APIs.
+It would also allow an easy reuse of existing Jakarta EE WebSocket handlers.
+
+## Note on *Partial* vs. *Whole* messages handlers
+
+The Jakarta WebSocket specification allows two kinds of message handlers which are described as:
+* *Whole* message handlers are notified by the container on arrival of a complete message (all frames recombined).
+* *Partial* message handlers are notified by the implementation when it is ready to deliver frames of a whole message.
+
+As it is explicitly stated in the JavaDoc part of the specification, the *partial* message handler is implementation-dependent.
```
laeubi, line 57: This is unclear, the difference between partial and whole is if the message is complete or not, while handling partly messages might be dependent on the protocol implementation it does not mean it is specific to the Webservice API implementation.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+
+### UC1: Define and register a WebSocket server endpoint using annotations
+
+The code below is a WebSocket handler that we would want to register on the `/ws/work/{targetId}` endpoint.
+
+The OSGi WebSocket Whiteboard implementation is responsible to register an instance of this service as WebSocket endpoint and message handler, either directly or using a proxy endpoint class.
+
+```java
+@Component(service = WebSocketEndpointExample.class, scope = ServiceScope.PROTOTYPE)
+@WebSocketWhiteboardServletPattern("/ws/work/{targetId}") // See "WebSocket pattern description"
```
laeubi, line 70: This seems missing a `@ServerEndpoint` (what would carry the path already)

If an implementaion is interfacing with the WebSocket Spec one *might* want to resuse annotation and/or properties from the Servlet spec but having it explicit here seem not useful

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+        // Process pong message
+    }
+}
+```
+
+#### WebSocket pattern description
+
+In a Jakarta environment, we would annotate this class with `@ServerEndpoint("/ws/work/{targetId}")` and register it in a Jakarta WebSocket-ready HTTP server.
+
+In OSGi, we might want to use annotations similar to the ones in the Jarkata Servlet whiteboard specification to provide more control, *e.g.* `@WebSocketWhiteboardServletPattern("/ws/work/{targetId}")`.
```
laeubi, line 131: Why? While a servlet *might* be one way to implement this, it is not mandatory and the `@ServerEndoint` carry enough information already.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### WebSocket subprotocols
+
+The [WebSocket specification (RFC 6455)](https://datatracker.ietf.org/doc/html/rfc6455#section-1.9) indicates that clients can request, during the handshake, can give a list of subprotocols names they expect, in order of preference.
+If a list of subprotocols is specified, the server must use and return the subprotocol it selected from that list.
+
+RFC 6455 doesn't explicitly specifies what happens if the server doesn't support any of the client subprotocols.
+
+The list of subprotocols supported by the OSGi WebSocket Whiteboard server endpoint can be given using a service property of type `string[]`.
```
laeubi, line 140: Subprotocols are already configured using `@ServerEndpoint("/ws/work/{targetId}", subprotocols = {"a", "b", "c"}`

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+A class annotation, *e.g.* `WebSocketWhiteboardSubprotocols`, could be used to ease the declaration of such list.
+
+#### Service scope
+
+The Jakarta WebSocket specification expects one instance of WebSocket handler per session.
```
laeubi, line 145: This is not true. The Specification states that a server is usually one instance, a client is an instance per connection if not otherwise configured.

Servers *must* be able to handle different client request with different sessions!

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+As a result, it is expected by the OSGi WebSocket Whiteboard specification to consume services with the `PROTOTYPE` service scope, to obtain a new instance of the service per session.
+
+Implementations using the `SINGLETON` service scope must manage thread safety and session handling by themselves.
+The `BUNDLE` service scope should be considered as a `SINGLETON`.
+
+### UC2: Define and register a WebSocket server endpoint using the `Endpoint` class
+
+This is the equivalent of [UC1](#uc1-define-and-register-a-websocket-server-endpoint-using-annotations) but extending a class instead of using annotations.
```
laeubi, line 153: Same as above, there must only one instance per server endpoint! In the sense of the Whiteboard pattern I won't support programatic endpoints at al as annotation based ones already supply what needed here.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+        // Error handling
+    }
+}
+```
+
+The OSGi WebSocket Whiteboard implementation is responsible to register an instance of this service as WebSocket endpoint, either directly or using a proxy endpoint class.
+
+The message handler registration is done by the endpoint in the `onOpen` method.
+
+Like the handler of [UC1](#uc1-define-and-register-a-websocket-server-endpoint-using-annotations), it is expected that the endpoint service has a `PROTOTYPE` scope.
```
laeubi, line 183: See above, this is not mandatory.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### Lack of URL parameters
+
+Extending the `Endpoint` class forbids to have extra parameters to the handling methods.
+As a result, it is not possible to have access to the URL parameters defined in the servlet pattern in that use case.
```
laeubi, line 188: This has nothing to do with servlet.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+  * `byte[]`
+  * `InputStream`
+  * `Decoder.Binary`
+  * `Decoder.BinaryStream`
+* For pong messages:
+  * `PongMessage`
+
+This interface lacks the `Session` argument, which is required to send data back to the peer.
+As a result, it might be preferred to define an OSGi specific interfaces that adds this argument and indicates the explicit kind of handled WebSocket frame.
+The OSGi WebSocket Whiteboard implementation will have to give the `Session` instance it obtains in the proxy message handler it registered.
```
laeubi, line 212: Message handlers are very restrictive on jakarta websocket spec, it seems not suitable to have any complex code in them. As per specification handlers are exspected to be added (only one handler per type!) to the session therefore I would not put Listeners in the responsibility of the Whiteboard Service.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+		public void onMessage(Session session, ByteBuffer message);
+	}
+
+	public interface Pong {
+		public void onMessage(Session session, PongMessage message);
+	}
+}
+```
+
+The OSGi WebSocket Whiteboard implementation is responsible to register an instance of this service using a basic WebSocket endpoint.
```
laeubi, line 230: See above should not be the responsibility of the Whiteboard Extender

laeubi (commented):
I think the requirements document has some serious flaws, it is currently focused on servlet integration not on the websocket spec. It also seem to limit implementations more than necessary.

.design/requirements-575.md
```diff
@@ -0,0 +1,269 @@
...
+## Introduction
+
+The OSGi Compendium R6 introduced the HTTP Whiteboard specification, upgraded in R8.1 to the Whiteboard specification for Jakarta Servlet.
+It provides a light and convenient way of using servlets in a an OSGi environment through the use of the Whiteboard Pattern.
+
+WebSocket is communication protocol designed to work over the HTTP protocol.
+It enables bidirectional exchange of messages between the web client and the server.
+Like an HTTP servlet, a WebSocket handler is bound to a URL.
+
+The conversion from an HTTP connection to a WebSocket stream is done by sending a `GET` HTTP request with a set of headers requesting a connection upgrade.
```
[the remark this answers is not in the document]
stbischof: https://www.rfc-editor.org/rfc/rfc6455


1.1
```
The WebSocket Protocol is designed to supersede existing
   bidirectional communication technologies that use HTTP as a transport
   layer to benefit from existing infrastructure (proxies, filtering,
   authentication).  Such technologies were implemented as trade-offs
   between efficiency and reliability because HTTP was not initially
   meant to be used for bidirectional communication (see [[RFC6202](https://www.rfc-editor.org/rfc/rfc6202)] for
   further discussion).  The WebSocket Protocol attempts to address the
   goals of existing bidirectional HTTP technologies in the context of
   the existing HTTP infrastructure; as such, it is designed to work
   over HTTP ports 80 and 443 as well as to support HTTP proxies and
   intermediaries, even if this implies some complexity specific to the
   current environment.  However, the design does not limit WebSocket to
   HTTP, and future implementations could use a simpler handshake over a
   dedicated port without reinventing the entire protocol.  This last
   point is important because the traffic patterns of interactive
   messaging do not closely match standard HTTP traffic and can induce
   unusual loads on some components.
   ```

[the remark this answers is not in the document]
stbischof: just wanted to point to the position in original rfc

laeubi:
I have investigated a bit further and it seems there is a standard compliant way to process an `UPGRADE` request see [here](https://docs.oracle.com/javaee/7/tutorial/servlets014.htm) in the servlet spec.

So an implementation of the Jakarta Websockets can implement a `Filter` on a context path using servlet whiteboard (what then becomes the root of the Webservice path) and listen for upgrade request, initiate the described procedure and then take over the stream, scheduling the request for processing by the webservice streams.

So this can work here like for Jakara Webservice specification, where we recommended to integrate with the Servlet Whiteboard but do not mandate its usage. This of course does not prevent an implementation to implement both in one service if it is supported native by the underlying container.

laeubi:
Some code samples used in the spec can be found here as executable maven project:
- https://github.com/laeubi/osgi-websockets

laeubi:
Here is the current state as a PDF for review if anyone is interested in the current state: 
[osgi.cmpn-8.1.0.pdf](https://github.com/user-attachments/files/17649089/osgi.cmpn-8.1.0.pdf)

laeubi:
I have now completed a first draft of the specification ready for review (**linked in the inital PR text**), and completed the implementation (**linked in the inital PR text**) so it has reached a state where it can be used already, so I think it would be possible to start with writing some tests to further refine the spec. @timothyjward  @stbischof @tcalmant 

I'll now try to demonstrate integration with the Servlet Whiteboard, from the specification we have some choices:

- [ ] Make it mandatory to support Servlet Witheboard
- [ ] Make it optional but recommend to support Servlet Witheboard
- [ ] Leave it to the implementation
- [ ] Let a Runtime make something similar available like the [ServerContainer](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/websocket/server/servercontainer) available in the service registry so it is possible to implement such support independent from the Websocket itself. using a method like `ServerContainer.upgradeHttpToWebSocket(Object, Object, ServerEndpointConfig, Map<String, String>)`.

laeubi:
> This way you may be able to Update to the beta Version

I'll give it a try!

PR #751, opened 2024-09-14
add clarification on same package import & export

chrisrueger:
add note about possible side effects if resolver choses "External"

TODO add more info


stbischof:
I really like the idea to show in examples what problems could happen, and what advantages and disadvantages several approaches of imports and Exports and embedding APIs or not can have.

But should we really add recommendation to handle Exports and imports like this in the spec.?


@tjwatson 
@timothyjward 
@pkriens 
@bosschaert

chrisrueger:
> But should we really add recommendation to handle Exports and imports like this in the spec.?

Good question and I am very open to suggestions. I just followed a proposal from a slack chat to suggest a spec adddition here like I did. But I am fine with any other suggestion. I just simply would like to have an "official" place (where I can link to from other places) for stuff like this (given the information I added is correct.. what is still to be determined by the group).

laeubi:
> But should we really add recommendation to handle Exports and imports like this in the spec.?

As an example we have similar for Require Bundle here:

https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3198951

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2231,10 +2231,58 @@ package-names   ::= package-name                // See <xref
...
       to its compatibility requirements, being either a consumer or a provider
       of that API. See <xref linkend="framework.module-semantic.versioning"/>
       for more information.</para>
+
+      <para>Also see the section about <xref linkend="framework.module-import.export.same.package"/> 
+      for more information about how the resolver handles this and what problems could arise with substitution.</para>
+
+      <note>
+      <para>
+		  As a recommendation: <emphasis role="bold">Library bundles</emphasis>
```
laeubi, line 2244: I find this a bit superfluous, especially as there is no definition of "library bundle" anywhere in the spec. Also it implies it should be used for "other bundles" I simply would not suggest using it at all, for me it would be enough to state that it is possible (e.g. not disallowed!) to import/export the same package.

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -3110,10 +3158,22 @@ B: Bundle-SymbolicName: B
...
       <code>osgi.wiring.package</code>, a requirement definition of a bundle
       may be wired to a capability definition of that same bundle.</para>
+
+      <para>
+	  	<emphasis role="bold">Notes on <emphasis>External</emphasis></emphasis>: 
+      If the resolver choses to resolve to an export statement in another bundle 
+		  the framework must <emphasis>discard</emphasis> the export package
+		  of the current bundle. This can lead to cases where consumers 
+		  for a version of the current bundle become <emphasis>unresolvable</emphasis>. 
+		  This effect can happen if the import has <emphasis>no</emphasis> version range
```
laeubi, line 3172: Actually this can happen always, it only depends on the circumstances and used import/export ranges.

laeubi (commented):
I think the previous paragraphs are better rewritten instead of adding to verbose explanation afterwards.

e.g. It is unclear why import+export improves collaboration, this is questionable as they can simply collaborate by not embedding API in a much better way.

Also the section that recommends substitution should be replaced with an explanation that it is *not* recommended and only useful in very rare cases!

chrisrueger:
> e.g. It is unclear why import+export improves collaboration, this is questionable as they can simply collaborate by not embedding API in a much better way.

How would you explain it?

Here is my attempt:

> Bundles willing to import exported packages allow the framework to _substitute_ exports for imports and improve collaboration, because the framework can try to avoid separate class-spaces.

Not optimal. I tried to somehow use this **blog post** but could not really come up with a easier explanation without giving a  huge example.

osgi.specs/docbook/core/003/framework.module.xml
chrisrueger: > I think the previous paragraphs are better rewritten instead of adding to verbose explanation afterwards.

I have pushed a rewrite of the previous paragraph and incorporated some parts of my notes in it. This reduced the size of my additional note.
I rephrased the existing section to now emphasize that substitution is not recommended and only useful in rare cases. Let's see how that works.

> I find this a bit superfluous, especially as there is no definition of "library bundle" anywhere in the spec. 

I rewrote this part. In the previous version of the spec the word "libraries" was (and is still ) used though. I rephrased it to "general/purpose java libraries.

osgi.specs/docbook/core/003/framework.module.xml
chrisrueger: I removed the last sentence.

> This effect can happen if the import has <emphasis>no</emphasis> version range...

Better?

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2198,18 +2198,29 @@ package-names   ::= package-name                // See <xref
...
-      <emphasis>substitute</emphasis> exports for imports.</para>
-
-      <para>Though substitution is recommended to increase collaboration, it
-      is not always possible. Importing exported packages can only work when
-      those packages are pure API and not encumbered with implementation
-      details. Import of exported packages should only be done when:</para>
+      <para>Bundles that collaborate require the same class loader 
+      for the types used in their interaction. 
+      When multiple bundles export the same package containing collaboration types, 
+      they may end up in separate class-spaces, which can hinder collaboration.
```
timothyjward, line 2204:
```suggestion
      they will end up in separate class-spaces, which prevents those collaboration types being shared between those bundles. 
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2198,18 +2198,29 @@ package-names   ::= package-name                // See <xref
...
-      is not always possible. Importing exported packages can only work when
-      those packages are pure API and not encumbered with implementation
-      details. Import of exported packages should only be done when:</para>
+      <para>Bundles that collaborate require the same class loader 
+      for the types used in their interaction. 
+      When multiple bundles export the same package containing collaboration types, 
+      they may end up in separate class-spaces, which can hinder collaboration. 
+      Bundles willing to import exported packages allow the framework to
+      <emphasis>substitute</emphasis> exports for imports and increase collaboration,
+      by trying to avoid separate class-spaces.
```
timothyjward, line 2207:
```suggestion
      by trying to avoid separate class-spaces. This is achieved by <emphasis>disabling</emphasis> the export in one or more bundles, replacing it with an import instead. The resulting class-space is larger, increasing collaboration, because there are fewer class loaders loading the exported package.
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2198,18 +2198,29 @@ package-names   ::= package-name                // See <xref
...
+      </para>
+
+      <para>
+      The use of substitutable exports had historical advantages:
+      Implementation bundles embedded and provided the
```
timothyjward, line 2212:
```suggestion
      Implementation bundles typically embedded and provided the 
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2198,18 +2198,29 @@ package-names   ::= package-name                // See <xref
...
+      they may end up in separate class-spaces, which can hinder collaboration. 
+      Bundles willing to import exported packages allow the framework to
+      <emphasis>substitute</emphasis> exports for imports and increase collaboration,
+      by trying to avoid separate class-spaces.
+      </para>
+
+      <para>
+      The use of substitutable exports had historical advantages:
+      Implementation bundles embedded and provided the 
+      API packages by OSGi specifications that they implement,
```
timothyjward, line 2213:
```suggestion
      API packages of the OSGi specifications that they implemented
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2198,18 +2198,29 @@ package-names   ::= package-name                // See <xref
...
+      <emphasis>substitute</emphasis> exports for imports and increase collaboration,
+      by trying to avoid separate class-spaces.
+      </para>
+
+      <para>
+      The use of substitutable exports had historical advantages:
+      Implementation bundles embedded and provided the 
+      API packages by OSGi specifications that they implement, 
+      to make themselves easier to deploy. 
+      Instead going forward you should consider not embedding API packages and declare them as dependencies
```
timothyjward, line 2215:
```suggestion
      Modern deployment tools mean that deploying bundles with separate API is now far simpler. Therefore you should now consider <emphasis>not</emphasis> embedding API packages and instead consume them
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2198,18 +2198,29 @@ package-names   ::= package-name                // See <xref
...
+      </para>
+
+      <para>
+      The use of substitutable exports had historical advantages:
+      Implementation bundles embedded and provided the 
+      API packages by OSGi specifications that they implement, 
+      to make themselves easier to deploy. 
+      Instead going forward you should consider not embedding API packages and declare them as dependencies 
+      using valid <code>Import-Package</code> manifest headers with a version range. 
+      At runtime, simply use the API runtime bundles provided by OSGi.
```
timothyjward, line 2217: It's not just OSGi specifications - we need to be sure people know that it applies to their own code too.

```suggestion
      At runtime the API can be provided by an API bundle, such as the specification API bundles provided by OSGi. 
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      for more information about how the resolver handles this and what problems could arise with substitution.</para>
+
+      <note>
+      <para>
+		  Bundles should only import such an exported package if they are willing to get that package 
+		  from another (unknown) bundle and omit their own export from the framework.
+      </para>
+      
+      <para>
+      Multiple bundles provide the same API packages
```
timothyjward, line 2263:
```suggestion
      Multiple bundles providing the same API packages 
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      <note>
+      <para>
+		  Bundles should only import such an exported package if they are willing to get that package 
+		  from another (unknown) bundle and omit their own export from the framework.
+      </para>
+      
+      <para>
+      Multiple bundles provide the same API packages 
+      can lead to conflicts in non-OSGi environments.
+      But also in OSGi environments resolution conflicts can happen
```
timothyjward, line 2265:
```suggestion
      In OSGi environments these scenarios can often be supported, but resolution conflicts can still occur
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      
+      <para>
+      Multiple bundles provide the same API packages 
+      can lead to conflicts in non-OSGi environments.
+      But also in OSGi environments resolution conflicts can happen 
+      when multiple versions of the same bundle are present (see <xref linkend="framework.module-import.export.same.package"/>).
+      </para>
+      
+      <para>
+      If you need do embed the API packages in an implementation bundle
```
timothyjward, line 2270:
```suggestion
      If you need to embed exported API packages in an implementation bundle 
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      Multiple bundles provide the same API packages 
+      can lead to conflicts in non-OSGi environments.
+      But also in OSGi environments resolution conflicts can happen 
+      when multiple versions of the same bundle are present (see <xref linkend="framework.module-import.export.same.package"/>).
+      </para>
+      
+      <para>
+      If you need do embed the API packages in an implementation bundle 
+      then be sure to only embed the API packages you directly implement.  
+      In most all cases you should not embed the APIs that you use but do not implement.
```
timothyjward, line 2272:
```suggestion
      You should not embed and export APIs that you use but do not implement.  
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      when multiple versions of the same bundle are present (see <xref linkend="framework.module-import.export.same.package"/>).
+      </para>
+      
+      <para>
+      If you need do embed the API packages in an implementation bundle 
+      then be sure to only embed the API packages you directly implement.  
+      In most all cases you should not embed the APIs that you use but do not implement.  
+      For example, if you use the <emphasis>OSGi Log Service</emphasis> you should avoid 
+      embedding the <code>org.osgi.service.log</code> package 
+      unless your bundle is actually providing the implementation of the Log Service.
```
timothyjward, line 2275:
```suggestion
      unless your bundle is actually providing an implementation of the Configuration Admin Service.
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      can lead to conflicts in non-OSGi environments.
+      But also in OSGi environments resolution conflicts can happen 
+      when multiple versions of the same bundle are present (see <xref linkend="framework.module-import.export.same.package"/>).
+      </para>
+      
+      <para>
+      If you need do embed the API packages in an implementation bundle 
+      then be sure to only embed the API packages you directly implement.  
+      In most all cases you should not embed the APIs that you use but do not implement.  
+      For example, if you use the <emphasis>OSGi Log Service</emphasis> you should avoid
```
timothyjward, line 2273: The Log service is a bad choice as the framework provides it.

```suggestion
      For example, if you use the <emphasis>OSGi Configuration Admin Service</emphasis> you should avoid 
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      But also in OSGi environments resolution conflicts can happen 
+      when multiple versions of the same bundle are present (see <xref linkend="framework.module-import.export.same.package"/>).
+      </para>
+      
+      <para>
+      If you need do embed the API packages in an implementation bundle 
+      then be sure to only embed the API packages you directly implement.  
+      In most all cases you should not embed the APIs that you use but do not implement.  
+      For example, if you use the <emphasis>OSGi Log Service</emphasis> you should avoid 
+      embedding the <code>org.osgi.service.log</code> package
```
timothyjward, line 2274:
```suggestion
      embedding the <code>org.osgi.service.cm</code> package 
```

chrisrueger:
Thanks a lot @timothyjward for your review and suggestions. I committed them all.

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2198,18 +2198,29 @@ package-names   ::= package-name                // See <xref
...
+      by trying to avoid separate class-spaces. This is achieved by <emphasis>disabling</emphasis> the export in one or more bundles, replacing it with an import instead. The resulting class-space is larger, increasing collaboration, because there are fewer class loaders loading the exported package.
+      </para>
+
+      <para>
+      The use of substitutable exports had historical advantages:
+      Implementation bundles typically embedded and provided the 
+      API packages of the OSGi specifications that they implemented
+      to make themselves easier to deploy. 
+      Modern deployment tools mean that deploying bundles with separate API is now far simpler. Therefore you should now consider <emphasis>not</emphasis> embedding API packages and instead consume them
+      using valid <code>Import-Package</code> manifest headers with a version range.
```
laeubi, line 2216:
```suggestion
      using a <code>Import-Package</code> manifest headers with a proper version range (either provider or consumer). 
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
       clean API-implementation separation. OSGi services are carefully
       designed to be as standalone as possible. Many libraries intertwine API
       and implementation in the same package making it impossible to
-      substitute the API packages.</para>
+      substitute the API packages.
+      Developers of general-purpose Java libraries that include OSGi metadata, 
+      but were not originally designed for OSGi, should avoid importing exported packages.
```
laeubi, line 2245:
```suggestion
      but were not originally designed with OSGi in mind, should generally avoid importing exported packages.
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      In OSGi environments these scenarios can often be supported, but resolution conflicts can still occur
+      when multiple versions of the same bundle are present (see <xref linkend="framework.module-import.export.same.package"/>).
+      </para>
+      
+      <para>
+      If you need to embed exported API packages in an implementation bundle 
+      then be sure to only embed the API packages you directly implement.  
+      You should not embed and export APIs that you use but do not implement.  
+      For example, if you use the <emphasis>OSGi Configuration Admin Service</emphasis> you should avoid 
+      embedding the <code>org.osgi.service.cm</code> package
```
laeubi, line 2274: I'm not fully sure about the message of this section, obviously a package can contain classes that are implemented and some that are not, from a substitution POV it also actually don't matter (given one follow the rules of proper imports either consumer or provider).

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+
+      <note>
+      <para>
+		  Bundles should only import such an exported package if they are willing to get that package 
+		  from another (unknown) bundle and omit their own export from the framework. By definition, this means that a bundle with a subsitutable export may never load that package, rendering its class files and resources invisible to that bundle's class loader.
+      </para>
+      
+      <para>
+      Multiple bundles providing the same API packages 
+      can lead to conflicts in non-OSGi environments.
```
laeubi, line 2264:
```suggestion
      can lead to conflicts in when used outside OSGi environments (e.g. JPMS) and therefore might hinder further use of such bundles.
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2229,12 +2240,44 @@ package-names   ::= package-name                // See <xref
...
+      In OSGi environments these scenarios can often be supported, but resolution conflicts can still occur
+      when multiple versions of the same bundle are present (see <xref linkend="framework.module-import.export.same.package"/>).
```
laeubi, line 2266:
```suggestion
      Even though OSGi supports these scenarios, it might lead to resolution conflicts can when multiple versions of the same bundle are present (see <xref linkend="framework.module-import.export.same.package"/>) and consumers are having tight version range constraints or large dependency chains.
```

osgi.specs/docbook/core/003/framework.module.xml
```diff
@@ -2198,18 +2198,29 @@ package-names   ::= package-name                // See <xref
...
-      <emphasis>substitute</emphasis> exports for imports.</para>
-
-      <para>Though substitution is recommended to increase collaboration, it
-      is not always possible. Importing exported packages can only work when
-      those packages are pure API and not encumbered with implementation
-      details. Import of exported packages should only be done when:</para>
+      <para>Bundles that collaborate require the same class loader 
+      for the types used in their interaction. 
+      When multiple bundles export the same package containing collaboration types, 
+      they will end up in separate class-spaces, which prevents those collaboration types being shared between those bundles.
```
laeubi, line 2204:
```suggestion
      they will end up in separate class-spaces, which prevents those collaboration types being shared between those bundles or their consumers. 
```

osgi.specs/docbook/core/003/framework.module.xml
chrisrueger: Maybe @timothyjward has an opinion about this how to improve this section?

chrisrueger:
@laeubi thanks for your suggestions, I added them.

laeubi:
This one here gave another (interesting) example that is not really covered by the spec:

- https://github.com/bndtools/bnd/issues/6431#issuecomment-2609087792

From the spec text it is not clear if a substitution packages is only valid if it is _mandatory_ or if also _optional_ is allowed.

According to the rules to apply [here](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module-import.export.same.package):

> If a bundle has both import and export definitions for the same package, then the Framework needs to decide which to choose.
> 
> It must first try to resolve the overlapping import definition. The following outcomes are possible:
> 
> - External - If this resolves to an export statement in another bundle, then the overlapping export definition in this bundle is discarded.
> - Internal - If it is resolved to an export statement in this bundle, then the overlapping import definition in this bundle is discarded.
> - Unresolved - There is no matching export definition. In this case the framework is free to discard either the overlapping export definition or overlapping import definition in this bundle. If the export definition is discarded and the import definition is not optional then the bundle will fail to resolve.
> 
> The above only applies to the import and export package definitions of a bundle. For namespaces other than osgi.wiring.package, a requirement definition of a bundle may be wired to a capability definition of that same bundle. 

So _optional_ is valid and only applies in the last case, here but this was not very clear to me and I think adding a note there and linking to the section of the spec would be useful.

timothyjward:
> From the spec text it is not clear if a substitution packages is only valid if it is mandatory or if also optional is allowed.

>> If the export definition is discarded and the import definition is not optional then the bundle will fail to resolve.

> I think adding a note there and linking to the section of the spec would be useful.

I'm not sure why we would add a note. Optional is explicitly mentioned, as is what will happen with an optional import

laeubi:
This section does not talk about optional at all:

[3.6.6 Importing Exported Packages ](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3238802)

so currently the information is hidden in another chapter about resolving here

[3.8.1 Importing and Exporting the Same Package](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module-import.export.same.package)

timothyjward:
> This section does not talk about optional at all:
>[3.6.6 Importing Exported Packages](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3238802)

This explains why you might want to have a substitutable export. If we were to add anything here I would suggest it said 

_"When importing an exported package you should not make your `Import-Package` statement optional. As your bundle is exporting the package there is no possibility of it not being available at runtime, meaning that making the import optional serves no purpose."_.

timothyjward:
> Your exported package version range can't match the import range
This is already pretty broken!

> In all cases one can't really use the package internally or only reflectively or only with things available in all versions in wich case one could have used a wider import range (without optional) in the first place.

I would strongly argue that this would be a case where you needed two bundles with "normal" dependencies, and if they needed to share anything between themselves then they must do so by using a different common package (in a shared class space). Trying to push everything into one bundle and coping with multiple versions of the package is basically avoiding any help OSGi can give you and trying to manage dependency hell inside your bundle!

laeubi:
> This is already pretty broken!

I hoped so but the spec specifically allows it here [3.8.1 Importing and Exporting the Same Package](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module-import.export.same.package)

> Unresolved - **There is no matching export definition**. In this case the framework is free to discard either the overlapping export definition or overlapping import definition in this bundle. If the export definition is discarded and the import definition is not optional then the bundle will fail to resolve. 

Sounds not very useful but is allowed and resolver must handle it (one way or the other).

chrisrueger:
> This explains why you might want to have a substitutable export. If we were to add anything here I would suggest it said
> 
> _"When importing an exported package you should not make your `Import-Package` statement optional. As your bundle is exporting the package there is no possibility of it not being available at runtime, meaning that making the import optional serves no purpose."_.

@timothyjward @laeubi I added this suggestion in 77a0abe

Here is a preview of the section with the changes of this PR:

image

image

chrisrueger:
@stbischof I squashed all commits as discussed in the call
and also rebased on master

PR #774, opened 2024-10-25
Design/552 - JPA to Jakarta Persistence

maho7791:
First move from JPA tp Jakarta Persistence

- osgi.service
- spec chapter


timothyjward:
The first commit claims to be a “move” but isn’t a rename in Git. We should endeavour to keep the history links where we can, for example in [this commit for the Servlet whiteboard](https://github.com/osgi/osgi/commit/48b23e721fcb66db02d4a2d7af5ce29ac0ef4636) where files were moved, not copied.

Issue #641, opened 2023-11-22
[ds annotation tck] ConstructorInjection test does not honor default values


laeubi:
[ConstructorInjection](https://github.com/osgi/osgi/blob/main/org.osgi.impl.bundle.component.annotations/src/org/osgi/impl/bundle/component/annotations/ConstructorInjection.java) testcase currently fails when one do not set the `field-collection-type` attribute with the following error:

```
java.lang.AssertionError: [reference[@name='multiple']/@field-collection-type for node testConstructorInjection] 
Expecting actual not to be null
	at org.osgi.test.support.xpath.AbstractXPathAssert.hasValue(AbstractXPathAssert.java:87)
	at org.osgi.test.cases.component.annotations.junit.DS14AnnotationsTestCase.testConstructorInjection(DS14AnnotationsTestCase.java:530)
```

[The spec says about this](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-field.injection):

> The type of objects set in the collection are specified by the field-collection-type attribute in the component description:
> -  service - The bound service object. **This is the default field collection type**.

Also [looking at the schema](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-schema) the attribute is marked as optional with a default of `service`

> ` <attribute name="field-collection-type" type="scr:Tfield-collection-type"  default="service" use="optional" />`

I therefore thing that here

https://github.com/osgi/osgi/blob/8ab4f46152fe43176b2a157f32e205a94487a481/org.osgi.test.cases.component.annotations/src/org/osgi/test/cases/component/annotations/junit/DS14AnnotationsTestCase.java#L530-L531

it must also accept the case that the value is not specified what will mean default attribute value is used.

stbischof:
Could you Pplease add a Test? here that covers this issue?

bjhargrave:
Yes, the test should be `hasOptionalValue` instead of `hasValue` to allow for the default value of `service`.

stbischof:
so please fixe a test or make it more distinct

laeubi:
@bjhargrave I mean there is no test that tests that the TCK itself is correct right?

bjhargrave:
> I mean there is no test that tests that the TCK itself is correct right?

That is a bit meta. We don't have tests to test the tests...

PR #780, opened 2024-11-09
field-collection-type is the default and must be optional

laeubi:
currently the test asserts that the value for field-collection-type is service, but as it is the default value it could be omitted.

Fix https://github.com/osgi/osgi/issues/641


stbischof:
Could you give me some references to spec text?

laeubi:
> Could you give me some references to spec text?

This is already described in the initial issue:
- https://github.com/osgi/osgi/issues/641

PR #781, opened 2024-11-09
Add testcase with empty array and a class in component property types

### LINKED ISSUE #646: [ds annotation tck] Usage of empty array and classes in componet property types is not covered by the TCK
Today I wanted to use the following component property type:

```
/**
 * Annotation that can be used for components to specify the provided adapter
 * types
 * 
 * @since 3.19
 */
@ComponentPropertyType
@Retention(RetentionPolicy.CLASS)
public @interface AdapterTypes {
	/**
	 * See {@link IAdapterFactory#SERVICE_PROPERTY_ADAPTABLE_CLASS}
	 * 
	 * @return the types that this class adapts from
	 */
	Class<?>[] adaptableClass();

	/**
	 * See {@link IAdapterFactory#SERVICE_PROPERTY_ADAPTER_NAMES}
	 * 
	 * @return the types that this class adapts to
	 */
	Class<?>[] adapterNames() default {};
}
```

sadly PDE-DS Generator generates wrong XML even though it passes the current TCK, so I suspect this case is currently missing from the TCK test cases.

laeubi:
Fix https://github.com/osgi/osgi/issues/646

### DIFF
--- a/org.osgi.impl.bundle.component.annotations/src/org/osgi/impl/bundle/component/annotations/ComponentPropertyTypesWithEmptyArray.java
+++ b/org.osgi.impl.bundle.component.annotations/src/org/osgi/impl/bundle/component/annotations/ComponentPropertyTypesWithEmptyArray.java
@@ -1,0 +1,34 @@
+/*******************************************************************************
+ * Copyright (c) Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


laeubi:
> Could you please add a reference to the Lines in Spec -text

See [here](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-component.property.mapping)

>> If the method type is Class or Class[], then the property type must be String or String[], respectively, whose values are fully qualified class names in the form returned by the Class.getName() method. 

It is not specifically explained how empty arrays must be handled, but as there is no way to define a "non-property" (it would still have the empty String as value) it is valid to assume that the case of an empty array should result in no property defined at all.

bjhargrave:
> @bjhargrave any opinion on that?

The TCK failed. So that should be fixed first.

bjhargrave (approved):
Seems to make sense and is already handled by Bnd as the component description generator.

bjhargrave:
@stbischof I will leave this to you to merge at your discretion.

stbischof:
@bjhargrave  thank you for the review

PR #787, opened 2024-11-17
Unregistering a ServiceRegistration must not throw IllegalStateException

tjwatson:
With PR #462 the unregister method relaxed the condition to throw an IllegalStateException if the ServiceRegistration was already unregistered. Therefore this test needs to be adjusted to not expext the exception.

### DIFF
--- a/org.osgi.test.cases.framework/src/org/osgi/test/cases/framework/junit/service/ServiceRegistryTests.java
+++ b/org.osgi.test.cases.framework/src/org/osgi/test/cases/framework/junit/service/ServiceRegistryTests.java
@@ -572,13 +572,6 @@ public class ServiceRegistryTests extends OSGiTestCase {
 		assertTrue(getContext().ungetService(reference2));
 
 		registration.unregister();
-		try {
-			registration.unregister();
-			fail("! Invalid, should have thrown IllegalStateException");
-		}
-		catch (IllegalStateException e) {
-			// ignore
-		}
 	}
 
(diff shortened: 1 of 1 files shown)


stbischof (approved):

stbischof:
@juergen-albert this is the Problem you had?

tjwatson:
> @juergen-albert this is the Problem you had?

I don't see how, unless you are referring to the original issue where we want to remove some of the conditions that cause an IllegalStateException.

I found this test while trying to implement the removal of the agreed IllegalStateExceptions in Equinox PR https://github.com/eclipse-equinox/equinox/pull/699

stbischof:
@tjwatson i refere to :
https://issues.apache.org/jira/plugins/servlet/mobile#issue/FELIX-6726

tjwatson:
> @tjwatson i refere to : https://issues.apache.org/jira/plugins/servlet/mobile#issue/FELIX-6726

No, I don't think this is the same issue problem that @juergen-albert ran into.  For that case the same service object was being passed to `ServiceObjects.ungetService` and the correct IllegalArgumentException was being thrown by the framework.  I think there was some confusion on what exception @juergen-albert was seeing because the title of the issue indicates IllegalStateException, but his comments in the issue indicate IllegalArgumentException.

I think Felix SCR should likely catch IllegalArgumentException there for cases where it may be called more than once for the same object when trying to clean up.

stbischof:
Thank you @tjwatson 

That helped me.

PR #800, opened 2025-01-21
setup test.cases.featurelauncher

stbischof:


stbischof:
@timothyjward you made a fix, but i cant see this. could you paste a link or push this into this branche

Issue #564, opened 2023-01-26
[core] effective:=active missing the the core spec?


laeubi:
https://docs.osgi.org/specification/osgi.core/8.0.0/framework.resource.html#d0e13546 says that the only "reserved" value is `resolve` but:

https://github.com/osgi/osgi/blob/19378f170291be6d36f02beb0f444940fe688340/org.osgi.resource/src/org/osgi/resource/Namespace.java#L141

indicates there is also another value `active` but I can't find it even here:

https://docs.osgi.org/specification/osgi.core/8.0.0/framework.namespaces.html

laeubi:
The resolver spec at laest mentions that there "might" be something like an "active" effectiveness:

https://docs.osgi.org/specification/osgi.core/8.0.0/service.resolver.html#d0e50547

tjwatson:
I don't recall why we added the `active` constant EFFECTIVE_ACTIVE to the `Namespace` class.  Usually I would argue such constants should be defined in another spec that defines their meaning.  The `active` effective value has no meaning to the core framework.

But since we do have this EFFECTIVE_ACTIVE constant, for https://docs.osgi.org/specification/osgi.core/8.0.0/framework.resource.html#d0e13546 we could add a sentence that refers to the `active` constant like this:

> For the OSGi Framework, the name `resolve` is reserved, this is also the default. Requirements and Capabilities may specify another value, such as `active`, but for the OSGi Framework the only value that makes the Requirement or Capability effective is the `resolve` value. The syntax is therefore:

rotty3000:
> For example, it could be possible to define an active time to resolve the service dependencies.

that text is non-normative. It's just an example, you could define any number of effectiveness values, but runtime only expects one single value `resolve`.

`active` is a bnd thing. Bnd is the only one who understands what that means and allows bnd to perform "assembly" operation.

rotty3000:
why it's a constants is also a mystery, but it's not part of the spec.

PR #803, opened 2025-01-28
[core] explain effective:=active value in the spec text

laeubi:
Currently Namespace has a constant EFFECTIVE_ACTIVE, but this is not referenced anywhere in the spec.

This now adds a little sentence at the spec text to make it appear as an example for other directives.

Fix https://github.com/osgi/osgi/issues/564


osgi.specs/docbook/core/006/framework.resource.xml
```diff
@@ -346,7 +346,11 @@
       that must match the <code>effective</code> directive.</para>
 
       <para>For the OSGi Framework, the name <code>resolve</code> is reserved,
-      this is also the default. The syntax is therefore:</para>
+      this is also the default. Requirements and Capabilities may specify another
+      value, such as <code>active</code> for requirements that must be present for
+      a bundle to become active executable (that is for example services it needs to consume)
+      but for the OSGi Framework the only value that makes the Requirement or Capability effective is the <code>resolve</code> value.
+      The syntax is therefore:</para>
```
tjwatson, line 353: This goes too far and implies the frame processes the requirement or capability when it is set to `active` before a bundle is allowed to become active.  Also, the term "active executable" is not used anywhere else in the specification (that I remember).  I'm not sure what it means.

I suggest we use what I recommended in the issue:

> For the OSGi Framework, the name resolve is reserved, this is also the default. Requirements and Capabilities may specify another value, such as active, but for the OSGi Framework the only value that makes the Requirement or Capability effective is the resolve value. The syntax is therefore:

tjwatson (changes requested):

timothyjward (approved):

PR #806, opened 2025-02-11
Enable maven as an alternative way to build osgi repository artifacts

laeubi:
Currently one can only build the sources with gradle, this enables to alternatively build them with maven and adds additional documentation to the CONTRIBUTING.md for users more familiar with that workflow.

See also
- https://tycho.eclipseprojects.io/doc/latest/BndBuild.html
- https://bnd.discourse.group/t/building-bnd-workspaces-with-pomless-maven-via-eclipse-tycho/494

FYI @chrisrueger


laeubi:
With the recent Tycho release everything now runs fine so this is ready to be submitted.

laeubi:
> I also have an full maven setup.

Then at best open a PR and maybe merge this one I'm fine with either way. The advantage for me on this setup is that it can work in parallel to the gradle build so users have a choice for local testing.

> JavaDoc Plugin is one of that.

Javadoc can be enabled but honestly it only wastes build times I never needed javadoc in the past 10 years since IDEs are good enough to generate the javadoc on the fly.

> I do not see a Action here until next release

Why not just merge it and let users working on the code have an alternative to gradle build (that often makes problem for me).

laeubi:
> I think that the point is that it's not actually simple. A maven build would need the ability to run the tests and generate the specs, and javadocj.

I would be happy to be even able to compile this stuff... generating the spec is something completely different and javadoc should already be supported by bnd anyways.

Still if it is desired one can iteratively add support for it and as described this does not _replace_ any existing tooling it just ads an additional opportunity.

> The OSGi build is very large, and has evolved over more than 20 years. The level of custom programming (first in ant, then in gradle) and the tight coupling to bnd workspaces/repositories makes it a very big job to move to maven.

This does completely support bnd workspace model with just a few simple lines... again no one claimed it replaces gradle build (even though it could)

>  @stbischof has looked as this, as have @juergen-albert and @maho7791 

I always found the bnd-maven- plugin much to complex, verbose and hard to use, that's why I came up with something that follows Tycho model to reuse existing configurations (just with bnd workspace model). So given that no of the other attempts has ever reached to the state of an PR it might be _good enough_ to be enabled in the meanwhile unless other approaches are ready...

> We would all love to see a move to maven, and have applied for funding to do it, but it will take someone a lot of work to get it done.

So here someone has done some work to move in that direction even for free ... from my perspective it does not help to claim "its a lot of work" one simply need to start somewhere!

PR #811, opened 2025-03-03
SpringSource EBR shutdown - remove repository

gg-dim:
Spring has shutdown the [SpringSource EBR on January 6, 2025](https://spring.io/blog/2024/12/02/repository-springsource-com-sunset). 
This PR should remove the repo and replace the 4 missing bundles:
* net.sourceforge.cglib:com.springsource.net.sf.cglib:2.2.0
* org.aopalliance:com.springsource.org.aopalliance:1.0.0
* org.apache.derby:com.springsource.org.apache.derby:10.4.1000003.648739
* org.apache.commons:com.springsource.org.apache.commons.logging:1.1.1


timothyjward:
This looks fine, but some of the version changes may have knock-on impacts in the TCKs. Hopefully the CI build will tell us whether it's all ok or not.

gg-dim:
I added the asm bundles after I have got dependency errors. 
```
org.osgi.framework.BundleException: Could not resolve module: org.osgi.impl.service.async [2]
  Unresolved requirement: Import-Package: org.apache.tools.ant; resolution:="optional"
  Unresolved requirement: Import-Package: org.apache.tools.ant.types; resolution:="optional"
  Unresolved requirement: Import-Package: org.objectweb.asm
```
My assumption was that the cglib version requires this. If you have a better solution/idea, please give me a hint.

timothyjward:
> My assumption was that the cglib version requires this. If you have a better solution/idea, please give me a hint.

That may be the case, but as far as I am aware the async service implementation doesn't use cglib...

gg-dim:
I take a look at the cglib/asm dependencies in the async service. The Enhancer from cglib is used in the AsyncService (line 135) to create a proxy and since I replaced the ```com.springsource.net.sf.cglib``` with a version 3.3 of the cglib, this brings the  asm dependencies.
As we will get problems with the cglib in Java 16+ we should replace it with byte-buddy or something else.
A good starting point could be https://mydailyjava.blogspot.com/2018/04/jdk-11-and-proxies-in-world-past.html?m=1

Issue #823, opened 2025-03-28
[tck][whiteboard] jakarta.servlet.error.exception_type attribute is assumed to be a String instead of Class


grgrzybek:
`org.osgi.test.cases.servlet.junit.ServletTestCase#test_table_140_4_HTTP_WHITEBOARD_SERVLET_ERROR_PAGE_exception()` test fails on Pax Web and I have exception:
```
java.lang.ClassCastException: class java.lang.Class cannot be cast to class java.lang.String (java.lang.Class and java.lang.String are in module java.base of loader 'bootstrap')
	at org.osgi.test.cases.servlet.junit.ServletTestCase$4BServlet.service(ServletTestCase.java:891) ~[?:?]
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:723) ~[jakarta.servlet-api-6.1.0.jar:6.1.0]
	at org.ops4j.pax.web.service.spi.servlet.OsgiInitializedServlet.service(OsgiInitializedServlet.java:118) ~[org.ops4j.pax.web.pax-web-spi-10.0.0.jar:?]
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:130) ~[org.ops4j.pax.web.pax-web-tomcat-10.0.0.jar:?]
	at org.ops4j.pax.web.service.spi.servlet.OsgiFilterChain.doFilter(OsgiFilterChain.java:114) ~[org.ops4j.pax.web.pax-web-spi-10.0.0.jar:?]
	at org.ops4j.pax.web.service.tomcat.internal.PaxWebStandardContext.lambda$createInitialOsgiFilter$1(PaxWebStandardContext.java:195) ~[org.ops4j.pax.web.pax-web-tomcat-10.0.0.jar:?]
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:109) ~[org.ops4j.pax.web.pax-web-tomcat-10.0.0.jar:?]
...
```

The problem is test's own Servlet `service()` implementation:
```java
@Override
protected void service(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
	invoked.set(true);
	String exception = (String) request.getAttribute(RequestDispatcher.ERROR_EXCEPTION_TYPE);
	response.getWriter().write((exception == null) ? "" : exception);
}
```

[...]

PR #830, opened 2025-05-08
Tck servlet improvements

grgrzybek:


cziegeler:
I dont understand why the servlet name needs to be set for those test.
If there is something to test with names, I suggest to create a new test, explicitly for that
The exception attribute change looks good

grgrzybek:
> I dont understand why the servlet name needs to be set for those test.

This is purely to make Pax Web pass these tests... That's why don't treat this as a must.

> If there is something to test with names, I suggest to create a new test, explicitly for that

I don't think there's a need for that... There's simply:

[OSGi CMPN Chapter 140 specification has](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.servlet.html#org.osgi.service.servlet.whiteboard.HttpWhiteboardConstants.HTTP_WHITEBOARD_SERVLET_NAME):

> Servlet names should be unique among all servlet services associated with a single [ServletContextHelper](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.servlet.html#org.osgi.service.servlet.context.ServletContextHelper).

Felix HTTP treats this "should" more loosely and Pax Web more strictly by not registering the servlet at all (internally it marks it as _disabled_ to be _enabled_ when the servlet with higher prioerity and the same name is unregistered). It's questionable which implementation adheres to `HttpWhiteboardConstants.HTTP_WHITEBOARD_SERVLET_NAME` _better_ :)

So it's not that TCK misses the check for servlet name uniqueness - simply CMPN Whiteboard spec doesn't say what to do when the conflict arises.

cziegeler:
Ok, if the spec is unclear, then the reverence implementation's behaviour can be seen as the clarification
Without your change, the servlets do not have a name and therefore do not fall under the mentioned constraint.

grgrzybek:
> Only _if_ the name property is set, the uniqueness constraint applies

I gently disagree... 

[140.4 Registering Servlets](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.servlet.html#service.servlet-i21223311):

> A servlet may be registered with the property `osgi.http.whiteboard.servlet.name` which can be used by servlet filters to address this servlet. If the servlet service does not have this property, the servlet name defaults to the fully qualified class name of the service object. 

So servlet _always_ have a name. And uniqueness applies here as well - this time implicitly.

[Servlet API 9.1. Obtaining a RequestDispatcher](https://jakarta.ee/specifications/servlet/6.1/jakarta-servlet-spec-6.1#obtaining-a-requestdispatcher):

> The `getNamedDispatcher` method takes a `String` argument indicating the name of a servlet known to the `ServletContext`. If a servlet is found, it is associated with a `RequestDispatcher` object and the object is returned. If no servlet is associated with the given name, the method must return `null`.

So combining these two you should get one servlet. And sure - you should get one with lower `service.id`. But it's even more implicit.

I agree that Felix HTTP is supposed to clear the vague parts of the spec and I'm not suggesting you should change TCK for Pax Web.
I can live with Pax Web not passing two tests ;)

One more argument from my side to add unique names is - I bet author of `org.osgi.test.cases.servlet.junit.ServletTestCase#test_140_4_16()` didn't _explicitly_ omit the names of the servlets (I may of course be wrong) - it just worked, so why worry? :)

Anyway - thanks for your time.

cziegeler:
Yes, every servlet has a name -  but not every servlet is *registered* with a name - that is a difference. In the latter case someone cares about the name, in the former no one cares about the name.
I was part of defining the spec and the way it is intended is exactly as I mentioned. If the name property is not set on a servlet, there can be as many servlets registered of the same class leading to the same name. Only if the name property is set, it will be made sure that just one servlet with that specified name is set in a given context.
If Pax Web decides to implement this differently I can guarantee you that some apps will not work with it.
Therefore changing the tests here will hide that an implementation deviates from the specification.

cziegeler:
and the same applies to filters

grgrzybek:
OK, I accept the explanation.
I may think about changing this behavior at some point. I've created ops4j/org.ops4j.pax.web#2048 to track this.

@cziegeler I'll change the PR to fix only #823, ok?

cziegeler:
Thanks, ok

cziegeler (approved):

grgrzybek:
For the record: https://issues.apache.org/jira/browse/FELIX-6777 is needed then.

stbischof:
@grgrzybek are you member of the Eclipse foundation?

stbischof:
Could you set you GitHub Account in the eclipse Account forms?

grgrzybek:
> Could you set you GitHub Account in the eclipse Account forms?

I set it today:
image

stbischof:
i revalidated eca check now green.

woulc you sign-off (-s)  the commit

grgrzybek:
I switched from using `-s` to `-S`, now pushing with both.

PR #839, opened 2025-07-04
Add support for Path bundle data

### LINKED ISSUE #824: [core] Support `Path` for bundle data
Currently there is only [`Bundle#getDataFile`](https://docs.osgi.org/javadoc/osgi.core/8.0.0/org/osgi/framework/Bundle.html#getDataFile-java.lang.String-) but this is bound to local file-systems.

It would be good do have `Path` support as well (e.g. `Bundle#getDataPath`) as this would allow storage of different kinds independent from a local file-system.

laeubi:
Fix #824


timothyjward:
> @timothyjward what is needed here to get this merged?

There's no block from me. @stbischof ?

PR #876, opened 2025-11-02
Revert "[ds] Add new component retention policy to control service lifecycle"

stbischof:
Reverts osgi/osgi#875


stbischof:
Sorry missclicked

PR #877, opened 2025-11-02
[ds] Add new component retention policy to control service lifecycle

### LINKED ISSUE #720: [ds] support a retention policy for components
Assuming a component at least provides one service, one currently has two options:

1. `immediate = true` what results in the component being activated and never deactivated
2. `immediate = false` what results in the component being activated when required and deactivated when use count drops to zero.

Now assume a component that is quite expensive to activate (or deactivate!) or retains a cache of such items, neither of both cases is really desirable:

- In the first case the component is always activated/deactivated even though it is probably never used.
- In the second case the component maybe gets activated/deactivated multiple times

The current workaround is to let the comment start tracking its own service in the activate method what is both tedious and ugly, because of this, I'd like to suggest adding a "component retention policy" that can have the value

- `DISCARD` (the default) that is the current behavior and deactivate/discard the component when its use count drops to zero
- `KEEP` when `immediate = false` is specified behaves like today, but do not deactivate/discard the component when its use count drops to zero (all other rules still apply e.g. if a required service goes away it is still deactivated).

This will now allow the component to be lazy activated but still kept around if for some time no one is using the service.

laeubi:
Fixes https://github.com/osgi/osgi/issues/720

---

Addresses #720: Components that are expensive to activate/deactivate or maintain caches currently have no control over lifecycle when service use count drops to zero. They either stay permanently active (`immediate=true`) or get deactivated on every idle period (`immediate=false`).

### Java API
- **`ComponentRetentionPolicy` enum**: `DISCARD` (default, current behavior) and `KEEP` (retain on idle)
- **`@Component.retentionPolicy()`**: New attribute controlling deactivation behavior when use count reaches zero

### Specification (service.component.xml)
- **Component Element**: Added `retention-policy` attribute documentation
- **Delayed Component lifecycle**: Updated to specify `KEEP` prevents deactivation on idle while still deactivating on unsatisfied conditions (required services unavailable, config deleted, component disabled, bundle stopped)

## Example Usage

```java
@Component(
    service = DatabaseService.class,
    immediate = false,
    retentionPolicy = ComponentRetentionPolicy.KEEP
)
public class DatabaseServiceImpl implements DatabaseService {
    private ConnectionPool pool;
    private Map<String, CachedResult> cache;
    
    @Activate
    void activate() {
        pool = new ConnectionPool();  // Expensive initialization
        cache = new HashMap<>();
    }
    
    // Component stays activated between service requests,
    // avoiding pool recreation and cache loss
}
```

### DIFF
--- a/org.osgi.impl.bundle.component.annotations/src/org/osgi/impl/bundle/component/annotations/HelloWorld16.java
+++ b/org.osgi.impl.bundle.component.annotations/src/org/osgi/impl/bundle/component/annotations/HelloWorld16.java
@@ -1,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0 
+ *******************************************************************************/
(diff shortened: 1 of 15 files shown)


laeubi:
Felix already support this with a custom extension, so it is more to make the felix feature officially specified here, see felix change:

- https://github.com/apache/felix-dev/pull/453

PR #879, opened 2025-11-03
Update Jakarta Web Services specification

laeubi:
This will address

- https://github.com/eclipse-osgi-technology/jakarta-webservices/issues/18
- https://github.com/eclipse-osgi-technology/jakarta-webservices/issues/9
- General improvements on the specification text itself

### DIFF
--- a/osgi.specs/docbook/160/service.jakartaws.xml
+++ b/osgi.specs/docbook/160/service.jakartaws.xml
@@ -39,14 +39,31 @@
     <title>Introduction</title>
 
     <para>
-    Webservices are an important part in todays maschine-to-maschine communication and the <xref linkend="service.jakartaws-i21217577"/> (formerly known as JAX-WS) specification offers a vendor neutral way of writing and using such services. 
-    It defines a standard Java-to-WSDL mapping which determines how web service operations are bound to Java methods when a SOAP message invokes an action.
-     
-    Managing endpoints in a dynamic way with integration to different listeners and transports can be a challenging task, this specification aims to make this task much more easier using the famous <xref linkend="service.jakartaws-whiteboard.pattern.ref" />
+    Web services are an important part in today's machine-to-machine communication. The <xref linkend="service.jakartaws-i21217577"/> (formerly known as JAX-WS) specification offers a vendor-neutral way of writing and using such services. 
+    It defines a standard Java-to-WSDL mapping (see <xref linkend="service.jakartaws-wsdl"/>) which determines how web service operations are bound to Java methods when a SOAP message (see <xref linkend="service.jakartaws-soap-protocol"/>) invokes an action.
+    </para>
+    
+    <para>
+    The Jakarta™ XML Web Services specification provides a comprehensive model for SOAP-based web services. It supports both document-oriented and RPC-oriented operations, 
+    handles SOAP 1.1 and SOAP 1.2 protocols, and provides a robust handler framework for message processing. 
+    The specification defines key concepts such as <emphasis>Service Endpoints</emphasis>, which are the server-side components that process web service requests (see <xref linkend="service.jakartaws-endpoint"/>),
+    and <emphasis>Handlers</emphasis>, which are processing units that can intercept and modify SOAP messages as they flow through the system (see <xref linkend="service.jakartaws-handler"/> and <xref linkend="service.jakartaws-handler-types"/>).
+    These handlers can be chained together to form a <emphasis>handler chain</emphasis>, allowing for modular cross-cutting concerns such as security, logging, or message transformation.
+    The order in which handlers process messages depends on the message direction, as described in <xref linkend="service.jakartaws-handler-execution"/>.
+    </para>
+    
+    <para>
+    This specification builds upon the concepts and terminology defined in the <xref linkend="service.jakartaws-i21217577"/>. 
+    Users and implementors of this specification are expected to be familiar with the Jakarta™ XML Web Services specification, 
+    as this whiteboard specification depends on those foundational concepts and does not redefine them.
+    </para>
+    
+    <para>
+    Managing endpoints in a dynamic way with integration to different listeners and transports can be a challenging task. This specification aims to make this task much easier using the well-known <xref linkend="service.jakartaws-whiteboard.pattern.ref" />
     and allow integration with existing transport specifications.
     
-    Even though different transports are supported, the most common one is the Http transport and also the only one that is mandatory for a <xref linkend="service.jakartaws-i21217577"/> implementation,
-    therefore this specification also focuses on the Http transport only and define a way to interact with it, implementations are free to support other kinds of transports as well.
+    Even though different transports are supported, the most common one is the HTTP transport and also the only one that is mandatory for a <xref linkend="service.jakartaws-i21217577"/> implementation.
+    Therefore, this specification also focuses on the HTTP transport only and defines a way to interact with it. Implementations are free to support other kinds of transports as well.
     </para>
 
     <para>The Whiteboard Specification for Jakarta™ XML Web Services supports:</para>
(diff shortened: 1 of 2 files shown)


osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -194,8 +216,23 @@ public class SOAPLoggingHandler implements SOAPHandler&lt;SOAPMessageContext&gt;
...
+</para>
+<para>
+ This handler chain is then passed to the Jakarta™ XML Web Services implementation. It is important to understand that the Jakarta™ XML Web Services runtime processes the handler chain differently depending on the message direction,
+ as specified in <xref linkend="service.jakartaws-handler-execution"/>:
+ <itemizedlist>
+   <listitem>
+     <para><emphasis>For outbound messages</emphasis>: Handler processing starts with the first handler in the chain and proceeds in the same order as the handler chain. This means the handler with the highest service ranking is processed first.</para>
+   </listitem>
+   <listitem>
+     <para><emphasis>For inbound messages</emphasis>: The order of processing is reversed. Processing starts with the last handler in the chain and proceeds in the reverse order of the handler chain. This means the handler with the lowest service ranking is processed first.</para>
```
timothyjward: I would put inbound first as it better matches the request/response flow, and say something like:

```
Inbound messages represent incoming Web Service requests, for example incoming HTTP requests from the internet. For inbound Messages the handlers are processed in ascending order, with lower ranked services executing first.
```

Then for outbound

```
Outbound messages represent outgoing Web Service responses, for example the result of a method call. For outbound messages the handlers are processed in descending order, with higher ranked services executing first.
```

Note that this is the exact reverse behaviour of the Servlet and Jakarta REST whiteboards, where incoming requests are processed by filters in descending order. We should discuss whether this is a good or bad thing.

osgi.specs/docbook/160/service.jakartaws.xml
```diff
@@ -194,8 +216,23 @@ public class SOAPLoggingHandler implements SOAPHandler&lt;SOAPMessageContext&gt;
     
 }</programlisting>
 <para>
- Handlers will be added to the Endpoint in service ranking order, that means that Handlers with a higher service ranking will be called before those with a lower service ranking.
- If two handlers have the same ranking they are sorted according to their service id, having handlers registered later being called later in the handler chain. 
+ Handlers are assembled into a handler chain in service ranking order. This means that Handlers with a higher service ranking will be positioned before those with a lower service ranking in the chain.
```
timothyjward: We need to discuss different types of handlers

```
When assembling a handler chain the implementation is required to group the handlers according to their types. Lower level, more protocol specific, handlers execute closer to the network layer and higher level handlers operate closer to the web service endpoint. For example with an incoming SOAP over HTTP request an `HttpHandler` would be called before a `SoapHandler`, and the `SoapHandler` would be called before a `LogicalHandler`.
Within a particular handler type the order of the handler chain is determined by service ranking order. This means that Handlers with a higher service ranking will be positioned earlier in the chain than those with a lower service ranking.
```

laeubi:
I have now clarified the specification in two points:

1. For "the user" we expect them to be familiar with the Jakarta Webservices specification and concepts (and especially implementers should)
2. Added a note about ordering across different types of handlers is not possible as they execute at different phases.

> Note that this is the exact reverse behavior of the Servlet and Jakarta REST whiteboards, where incoming requests are processed by filters in descending order (highest rank first). We should discuss whether this is a good or bad thing.

I one wants to change that, I think that should be brought up to the jakarta ws specification committee, the purpose of the whiteboard here is to make it integrate with the OSGi whiteboard concept not to make it "better" so I would always say we should adopt to what they propose.

From a server side this makes perfectly sense as we can assume that the inbound traffic for an implementation is mostly opaque but on the outbound you want to impose certain things like signing or encryption on the element level and so on. One can even easily register two handlers and place them at any end of the chain if the usecase require it.

laeubi:
> I’m simply saying that for all other similar OSGi integrations a higher service ranking causes your handler/filter to be executed earlier in the inbound chain, not later.

Because their specification they are foundation of are working that way, for example in Servlet filter spec there is literally no distinction between INBOUND and OUTBOUND your `Filter` are only called once and then you can maybe wrap the request/response and possibly invoke the next handler in the chain. In jakarta ws this is not the case, you operate on a message that is incoming, then others do that as well and then for the outbound the same

> I’m not sure why the Jakarta team would be involved in that decision at all. They don’t care about OSGi service ranking.

Because they have decided (and described) how handler chains work and what is applied in the order doing it the reverse in OSGi would be highly confusing because then it might feel more naturally for the INBOUND but is reversed (feels wrong) on the OUTBOUND. So anyone that is familiar with handlerchains from jakarta-ws would probably think it is a bug, then we must told them "No we did it differently because we feel the jakarta spec is doing it wrong'.
And I strongly believe we should not code against the spec here.

I'll try to mention this distinction in the spec, but it becomes way to verbose about the topic of handler ordering than it should in my opinion.

So apart from that thing, my question would be (and what was the actual intend in this regard) is it now clear enough for you to adjust the TCKs so we get passing tests?

osgi.specs/docbook/160/service.jakartaws.xml
laeubi: I hope it is now clear enough?

timothyjward:
> > I’m not sure why the Jakarta team would be involved in that decision at all. They don’t care about OSGi service ranking.
> 
> Because they have decided (and described) how handler chains work and what is applied in the order doing it the reverse in OSGi would be highly confusing because then it might feel more naturally for the INBOUND but is reversed (feels wrong) on the OUTBOUND. So anyone that is familiar with handlerchains from jakarta-ws would probably think it is a bug, then we must told them "No we did it differently because we feel the jakarta spec is doing it wrong'. And I strongly believe we should not code against the spec here.
> 

I think that I haven’t expressed myself clearly and that as a result you have misunderstood my concern.


The principle of these “integration” specifications is always fidelity first. Of course the outbound should be the reverse of the inbound. My point relates solely to the service ranking order, i.e. how the handler chain is defined. 

In both [Jakarta REST](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.jakartars.html#d0e132381) and [Servlet](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.http.whiteboard.html#d0e121055) whiteboards there is a clear explanation stating that Jakarta rules are followed, but in the event of ties service ranking is used. In both those cases the highest ranking service is called first on the inbound flow. This is the opposite of the web services whiteboard where the highest ranking service is called last on the inbound flow. This is purely an OSGi decision and does not impact the Jakarta rules at all.

laeubi:
> In both [Jakarta REST](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.jakartars.html#d0e132381) and [Servlet](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.http.whiteboard.html#d0e121055) whiteboards there is a clear explanation stating that Jakarta rules are followed, but in the event of ties service ranking is used.

- We don't have a ties here
- Handlers don't have a priority
- You always define a complete handler chain that is an ordered collection (an array) usually in the server configuration or in code
- We here use "service ranking order" what is the natural ordering for services in OSGi
- If we would "reverse" that only for trying to make one particular expectation mapping better it will be inconsistent to how handler chains are defined traditionally on jakarta-ws configurations.

So I think everything is consistent here and not really comparable to other specifications. Also other specifications does not matter in the sense as they are not used in conjunction so a SOAP handler will never be called from a rest endpoint so there is no risk of getting any inconsistencies (as we order services always consistently)

> In both those cases the highest ranking service is called first on the inbound flow.

That is not particular true. For JaxRS there are distinct handlers for inbound and outbound and you need to be careful if you implement both to do it on the right place. Also you said that service ranking is only used in the case of a ties.
For JakaratServlet there even is no distinction between inbound/outbound at all and we have also a filter chain approach where we uses *exactly* the same ordering as here, filters with a higher ranking are on the top of the chain.

> This is purely an OSGi decision and does not impact the Jakarta rules at all.

So my decision when writing the spec and writing the implementation was to be as close as to how one defines filter-chains in a traditional jakarta-ws (both in code and in xml configurations) and the TCK should reflect this.

laeubi:
I now added an execution flow diagram to the specification to make it hopefully more clear here:

- https://github.com/osgi/osgi/pull/879/commits/cdec7fc955267f76b70c7bf0da6f9a962ee69a27

timothyjward:
> The design branch is still messed up, lost all previous changes and is behind main
> 
> > This branch is [19 commits behind](https://github.com/osgi/osgi/compare/design/587...main) main.

I don't know the commit that would be considered "last good" so I'm unable to help. If it is commit cdec7fc955267f76b70c7bf0da6f9a962ee69a27 then this is still available in GitHub so I suggest that you work from there. The `main` branch is ongoing, so this branch will fall behind over time anyway.

PR #880, opened 2025-11-04
Use more recent gradle version

laeubi:
The current one only supports Java 19 but java 21 is current and 25 approaching ... also updates bnd from 6 > 7


laeubi:
@bjhargrave great, would you kindly merge this PR in the meanwhile?

PR #881, opened 2025-11-16
Add comprehensive examples for osgi.service and osgi.implementation namespaces

laeubi:
Currently the sections are quite bare and even though one usually should not write them manually its good to see how they would look like (see for example [here](https://github.com/eclipse-equinox/p2/pull/972)).

This now adds some more examples to make it easier to grasp how such headers look like.

### DIFF
--- a/osgi.specs/docbook/135/service.namespaces.xml
+++ b/osgi.specs/docbook/135/service.namespaces.xml
@@ -592,6 +592,33 @@ DynamicImport-Package:
       </tgroup>
     </table>
 
+    <para>For example, a bundle that provides the Log Service could declare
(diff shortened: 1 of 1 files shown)


osgi.specs/docbook/135/service.namespaces.xml
```diff
@@ -592,6 +592,33 @@ DynamicImport-Package:
       </tgroup>
     </table>
 
+    <para>For example, a bundle that provides the Log Service could declare
```
stbischof: `should`  woule be more clear that could

laeubi:
@stbischof @timothyjward what is needed to conclude on this?

Issue #372, opened 2021-10-15
core: Add support for ServiceLoader to the Core Framework


bjhargrave:
We have a Service Loader Mediator spec. But it is not built into the framework and currently depends upon weaving. 

I would like to explore adding support directly unto the framework for the Service Loader model. JPMS has direct support for the Service Loader model via `uses` and `provides` in the module-info file. JPMS has advantages since it is built into the java runtime, so we may not be able to do as well. But I suspect we can do better than the current Service Loader Mediator spec by building this function into the core framework.

maho7791:
I would support that as well, because the ServiceLoader Mediator requires its Capabilities to work properly. 
Especially Jarkarta API's (JPA, JAXB, Jax-Ws, Websockets) make use of the ServiceLoader to find the Implementation. The OSGi Manifests often event dont have the Capabilities or they are not complete. Having this support in the core framework, would solve major pain points.

laeubi:
> We have a Service Loader Mediator spec.

I once tried to use that but it seems not fitting very well because it seem to be limited to the use of [ServiceLoader.load() calls](https://blog.osgi.org/2013/02/javautilserviceloader-in-osgi.html) but I must confess I never wanted to use it that way but the other way round: Accessing  ServiceLoader specified services in an OSGi way

> But I suspect we can do better than the current Service Loader Mediator spec by building this function into the core framework.

I once written a extender-bundle that simply discovers SPI in `META-INF/services` and registered them as plain OSGi services that's quite useful and could even be made lazy when using a ServiceFactory.

> Especially Jarkarta API's (JPA, JAXB, Jax-Ws, Websockets) make use of the ServiceLoader to find the Implementation.

Jakarta can uses an [OSGi-aware serviceloader](https://github.com/eclipse-ee4j/glassfish-hk2-extra/blob/master/osgi-resource-locator/src/main/java/org/glassfish/hk2/osgiresourcelocator/ServiceLoader.java) that works quite well. I just always wondered why they did not simply delegate to standard java ServiceLoader.load() if nothing is found (it instead returns null then) this requires some care on the caller-side, but maybe this can server as an inspriration?

rotty3000:
[...]

PR #882, opened 2025-11-18
Enhance the specification for OSGi Core SPI support

laeubi:
as discussed in the issue:

- https://github.com/osgi/osgi/issues/372

Proof-of-conpect implementation:

- https://github.com/apache/felix-dev/pull/455
- https://github.com/eclipse-equinox/equinox/pull/853


laeubi:
I'll close this for now and start with a new PR targeting `main` for now. Then when we are ready to merge we can change the target branch this seems easier for me at the moment the `design/372` seems to be in a strange state some commits before but some behind and I can't see what was actually there.

So maybe rename the branch and create a fresh one then at this time.

PR #883, opened 2025-11-18
Add Logging Support to Resolver Service

laeubi:
## Overview

This PR enhances the OSGi Resolver Service Specification to provide optional logging support during resolve operations. This allows resolve contexts to gain visibility into the resolution process, which is invaluable for debugging complex dependency scenarios and understanding resolver behavior.

## Problem Statement

Currently, the Resolver Service operates as a "black box" during resolve operations. When resolution succeeds, callers receive a set of wires, and when it fails, they receive a `ResolutionException`. However, there is no way to observe:

- What the resolver is doing during the resolution process
- Which resources are being considered
- Why certain capability matches are preferred over others
- What causes the resolver to explore or reject different resolution paths
- Performance characteristics of the resolution

This lack of visibility makes it extremely difficult to:
- Debug complex resolution failures
- Understand why a particular resolution was chosen
- Optimize resolve context implementations
- Diagnose performance issues in large dependency graphs

## Solution

This PR adds an optional `getLogger()` method to the `ResolveContext` abstract class that returns an `Optional<Logger>`. This approach:

1. **Maintains backward compatibility** - The default implementation returns `Optional.empty()`, so existing resolve contexts continue to work without modification
2. **Uses existing OSGi Log Service** - Leverages the well-established `org.osgi.service.log.Logger` interface
3. **Is truly optional** - Resolver implementations are not required to call this method or use the logger
4. **Is non-intrusive** - Does not affect the resolver's behavior, only provides visibility

### 1. ResolveContext.java

Added a new method to the `ResolveContext` abstract class:

```java
/**
 * Returns an optional {@link Logger} that can be used during the resolve
 * operation to emit log messages.
 * <p>
 * The resolver may call this method to obtain a logger for emitting
 * diagnostic and informational messages during the resolve operation...
 * 
 * @return An {@link Optional} containing a {@link Logger} to be used during
 *         the resolve operation, or an empty Optional if no logger is
 *         available. Must not be {@code null}.
 * @since 1.2
 */
public Optional<Logger> getLogger() {
    return Optional.empty();
}
```

Key aspects:
- Returns `Optional<Logger>` to clearly indicate the logger may not be present
- Default implementation returns empty Optional for backward compatibility
- Comprehensive Javadoc explaining usage and expectations
- Marked with `@since 1.2` for version tracking

#### Updated Essentials Section
Added a new essential requirement:
- **Visibility** - Allow resolve contexts to receive logging information about the resolution process to aid debugging and diagnostics

#### New "Logging" Section
Added a comprehensive new section explaining:
- Purpose and benefits of resolver logging
- Types of information that can be logged
- How to provide a logger from a resolve context
- Thread-safety requirements
- Performance considerations
- Example implementation showing logger integration

#### Cross-References
Added proper DocBook cross-references to:
- `org.osgi.service.log.Logger` (hyperlink style)
- Log Service Specification (chapter 101)
- `ResolveContext.getLogger()` method (hyperlink style)

## Usage Example

Here's how a management agent can provide logging:

```java
import org.osgi.service.log.Logger;
import org.osgi.service.log.LoggerFactory;
import java.util.Optional;

public class MyResolveContext extends ResolveContext {
  private final Logger logger;

  public MyResolveContext(LoggerFactory loggerFactory) {
    this.logger = loggerFactory.getLogger(MyResolveContext.class);
  }

  @Override
  public Optional<Logger> getLogger() {
    return Optional.of(logger);
  }

  // ... other ResolveContext methods
}
```

The resolver implementation can then use this logger:

```java
public Map<Resource, List<Wire>> resolve(ResolveContext context) {
    Optional<Logger> logger = context.getLogger();
    
    logger.ifPresent(log -> log.debug("Starting resolution with {} mandatory resources", 
        context.getMandatoryResources().size()));
    
    // ... resolution logic ...
    
    logger.ifPresent(log -> log.info("Resolution completed successfully"));
    
    return result;
}
```

## Benefits

1. **Better Debugging** - Developers can see exactly what the resolver is doing and why certain decisions are made
2. **Performance Analysis** - Log messages can include timing information to identify bottlenecks
3. **Educational** - Helps developers understand how the resolver works
4. **Backward Compatible** - Existing code continues to work unchanged
5. **Standards-Based** - Uses the existing OSGi Log Service, no new dependencies
6. **Optional** - No performance impact if logging is not enabled

## Implementation Notes for Resolver Providers

Resolver implementations are encouraged but not required to:

1. Call `context.getLogger()` at the start of a resolve operation
2. Check `logger.isDebugEnabled()` before creating expensive log messages
3. Use appropriate log levels:
   - `TRACE` for detailed step-by-step information
   - `DEBUG` for general resolution progress
   - `INFO` for significant milestones
   - `WARN` for potential issues that don't prevent resolution
4. Return the same Logger instance if called multiple times during one operation
5. Ensure thread-safety when using the logger in multi-threaded resolution

## Testing

Note: As specified in the requirements, **no test cases were added**. The support for logging is completely optional:
- Resolver implementations are not required to call `getLogger()`
- Resolver implementations are not required to use the logger if provided
- Existing resolve contexts that don't override `getLogger()` continue to work

This makes the feature testable only through resolver-specific implementations that choose to support it.

## API Version

This change requires a minor version bump of the `org.osgi.service.resolver` package (to 1.2.0) due to the addition of a new method with a default implementation in the `ResolveContext` class.

## Related Specifications

- OSGi Log Service Specification (Chapter 101)
- OSGi Resolver Service Specification (Chapter 58)

## Backward Compatibility

✅ **Fully backward compatible**
- Default method implementation means existing subclasses don't need changes
- Optional return type clearly indicates the logger may not be present
- Resolver implementations that don't check for a logger continue to work
- No changes to existing method signatures or semantics

### DIFF
--- a/org.osgi.service.resolver/bnd.bnd
+++ b/org.osgi.service.resolver/bnd.bnd
@@ -6,4 +6,5 @@ Export-Package: ${p}.*; -split-package:=first
 -buildpath = \
     ${osgi.annotation.buildpath}, \
     org.osgi.framework;maven-scope=provided;version=1.8,\
-    org.osgi.resource;maven-scope=provided;version=1.0
+    org.osgi.resource;maven-scope=provided;version=1.0,\
+    org.osgi.service.log;maven-scope=provided;version=1.4
--- a/org.osgi.service.resolver/src/org/osgi/service/resolver/ResolveContext.java
+++ b/org.osgi.service.resolver/src/org/osgi/service/resolver/ResolveContext.java
@@ -37,7 +37,7 @@ import org.osgi.resource.Requirement;
 import org.osgi.resource.Resource;
 import org.osgi.resource.Wire;
 import org.osgi.resource.Wiring;
-import org.osgi.service.log.Logger;
+import org.osgi.service.log.FormatterLogger;
 
 /**
  * A resolve context provides resources, options and constraints to the
@@ -258,8 +258,8 @@ public abstract class ResolveContext {
 	}
 
 	/**
-	 * Returns an optional {@link Logger} that can be used during the resolve
-	 * operation to emit log messages.
+	 * Returns an optional {@link FormatterLogger} that can be used during the
+	 * resolve operation to emit log messages.
 	 * <p>
 	 * The resolver may call this method to obtain a logger for emitting
 	 * diagnostic and informational messages during the resolve operation. This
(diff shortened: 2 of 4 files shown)


laeubi:
As discussed last time I now

- Changed to `FormatterLogger` to make it explicit what kind of logger to expect
- Added a not about implication of required packages ranges

If anything is left from then please add a comment to the spec text for further iterations, or leave an approving review if you are fine with it in the current state.

PR #884, opened 2025-11-18
Enhance Resolver Service with SPI Support

laeubi:
Enables obtaining resolver instances outside OSGi frameworks using Java SPI, similar to `FrameworkFactory`

### Problem
Resolvers are needed in many scenarios before/outside OSGi:
- Framework initialization
- Build tools (Maven, Gradle, bnd)  
- IDEs analyzing dependencies
- Provisioning tools
- Feature launchers

Currently requires referencing specific implementation classes.

### Solution
New `ResolverFactory` interface using Java Service Provider mechanism, plus `osgi.resolver.class` framework property for resolver selection.

### Changes
- **ResolverFactory.java**: New interface for creating resolvers
- **Constants.java**: Added `FRAMEWORK_RESOLVER_CLASS` constant
- **service.resolver.xml**: New "Obtaining Resolvers Outside OSGi" section
- **framework.lifecycle.xml**: Added `osgi.resolver.class` property documentation

### Usage
```java
ServiceLoader<ResolverFactory> loader = 
    ServiceLoader.load(ResolverFactory.class);
Resolver resolver = loader.findFirst()
    .map(ResolverFactory::getResolver)
    .orElseThrow();
```

## Benefits
- Standard, discoverable way to obtain resolvers
- Frameworks expose resolvers for tooling
- Same resolution algorithm in dev and runtime
- Pluggable resolver implementations
- Optional logging for debugging


osgi.specs/docbook/core/058/service.resolver.xml
```diff
@@ -1415,6 +1416,144 @@ public List&lt;Capability&gt; findProviders( Requirement requirement) {
...
+        repositories to download and provision bundles often need resolution
+        capabilities but do not themselves run within OSGi.</para>
+      </listitem>
+
+      <listitem>
+        <para><emphasis>Feature Launchers</emphasis> - Tools that construct
+        and launch OSGi frameworks based on feature definitions (see <xref
+        linkend="service.featurelauncher"/>) need to resolve features before
+        the framework starts.</para>
+      </listitem>
```
tjwatson: I would avoid a reference to the feature launcher from the core specification.  I don't think it is necessary and the Feature Launcher is a specified provisioning tool so should be covered by the Provisioning Tool. I want to avoid links to other specifications that are not even published yet from the Core spec.

osgi.specs/docbook/core/058/service.resolver.xml
```diff
@@ -1415,6 +1416,144 @@ public List&lt;Capability&gt; findProviders( Requirement requirement) {
...
+
+      <listitem>
+        <para><emphasis>IDEs and Development Tools</emphasis> - Integrated
+        development environments need to analyze bundle dependencies and
+        predict resolution outcomes without running a full OSGi
+        framework.</para>
+      </listitem>
+
+      <listitem>
+        <para><emphasis>Provisioning Tools</emphasis> - Tools that operate on
```
tjwatson: I don't think we have used the term `Provisioning Tool` anywhere else in the specification.  I would use `Management Agent` instead:
```suggestion
        <para><emphasis>Management Agents</emphasis> - Agents that operate on
```

osgi.specs/docbook/core/058/service.resolver.xml
```diff
@@ -1415,6 +1416,144 @@ public List&lt;Capability&gt; findProviders( Requirement requirement) {
     </section>
   </section>
 
+  <section>
+    <title>Obtaining Resolvers Using SPI</title>
```
tjwatson: To be consistent with other core sections that specify `META-INF/services` resources.

```suggestion
    <title>Java Service Provider Configuration Support for Resolvers</title>
```

osgi.specs/docbook/core/058/service.resolver.xml
```diff
@@ -1415,6 +1416,144 @@ public List&lt;Capability&gt; findProviders( Requirement requirement) {
...
+
+      <para>With the contents:</para>
+
+      <programlisting># ACME Resolver Implementation
+com.acme.osgi.ResolverImpl</programlisting>
+
+      <para>The Resolver implementation class must have a public, no-argument
+      constructor. Each instance is independent and can be used to perform
+      resolution operations.</para>
+    </section>
```
tjwatson: I would model this after other sections where we spec requirements on a `META-INF/services` resource.  For example, like https://docs.osgi.org/specification/osgi.core/8.0.0/framework.connect.html#framework.connect.service.provider

tjwatson (changes requested):

laeubi:
@tjwatson I hopefully addressed all your concerns so this would be ready for a final review.

PR #885, opened 2025-11-19
We only want to test build!

laeubi:
Test the build


stbischof:
Lovely

laeubi:
@stbischof I now squashed and cleanup my experiment here, it should then be ready for submission

PR #889, opened 2025-11-20
Fix blind cast in ServletTestCase error handler

laeubi:
Add type checking for ERROR_EXCEPTION_TYPE attribute to handle both String and Class types properly. This fixes the ClassCastException that occurred when the servlet container provides the exception type as a String (which is the correct behavior per Jakarta Servlet spec) instead of blindly casting to Class.


laeubi:
@timothyjward what needs to be done for this to be merged?
Currently all PRs fail with this failure on `main`...

timothyjward (approved):

PR #903, opened 2026-01-05
Typed Event version 1.1 TCK

gg-dim:
I have started to extend the TCK for version 1.1. of the typed event service. Some tests are failing as the Apache Aries Typed Event implementation is not compatible in some details. The aries fixes will be done in the [aries project](https://github.com/apache/aries-typedevent).
Some other test are failing because isn't working properly.

### DIFF
--- a/org.osgi.service.typedevent/src/org/osgi/service/typedevent/TypedEventBus.java
+++ b/org.osgi.service.typedevent/src/org/osgi/service/typedevent/TypedEventBus.java
@@ -27,7 +27,7 @@ import org.osgi.annotation.versioning.ProviderType;
  * this service and call one of the event delivery methods.
  * 
  * @ThreadSafe
- * @author $Id$
+ * @author $Id: 8c3d712804626b7c14c5ec563fad9ddde3e1d13b $
  */
 @ProviderType
 public interface TypedEventBus {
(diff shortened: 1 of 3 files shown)


org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/EventHistoryTest.java
```diff
@@ -0,0 +1,342 @@
...
+							EventA.class.getName(),
+							TypedEventConstants.TYPED_EVENT_HISTORY, 1));
+
+			EventA newEvent = new EventA();
+			newEvent.a = "new";
+			eventBus.deliver(newEvent);
+
+			InOrder inOrder = inOrder(handler);
+			inOrder.verify(handler, timeout(1000)).notify(eq(TOPIC_A),
+					eq(historicalEvent));
```
timothyjward: This equal won't work as `EventA` doesn't have an equals method


```suggestion
					Mockito.argThat(evA -> "historical".equals(evA.a)));
```

org.osgi.test.cases.typedevent/src/org/osgi/test/cases/typedevent/junit/TypedEventMonitorConfigurationTest.java
```diff
@@ -0,0 +1,380 @@
...
+	@InjectService
+	TypedEventBus eventBus;
+
+	@InjectService
+	TypedEventMonitor monitor;
+
+	protected static final String TOPIC_A = EventA.class.getName()
+			.replace(".", "/");
+
+	@BeforeEach
```
timothyjward: This test has the same issue as `EventHistoryTest` - the reinit needs to target the bundles hosting the services, and the `eventBus` and `monitor` fields should be removed and supplied as parameters instead.

stbischof (approved):

Issue #688, opened 2024-03-01
[ds] allow a component to implement ServiceFactory / PrototypeServiceFactory


laeubi:
Currently a component must implement the services it exposes through the service factory.

There are sometimes cases where one want to expose a service with DS, but creation of such service won't be possible by DS what for example includes:

1. The service is not an interface but a final class that needs constructor arguments that are dependencies or need to be constructed from injected services, java records are a good example as well as many (non OSGi) libraries
2. The constructor is invisible and instances can only be created by a static factory

## Current state

This currently forces one to use some indirection, e.g have a component that does not expose a service, injects a bundle context and then register/unregistered the service. This contradicts the idea of letting the SCR manage the service lifecycle, for example this means such a component must always be active and possibly registering services that are never used.

### Proposed solution

If a component exposes **exactly one service `S`** instead of requiring the component to implement this service it is allowed to alternatively implement the `ServiceFactory` / `PrototypeServiceFactory` interface with the generic parameter of `<S>`.

stbischof:
Register manually the service on bundleContext.

PR #904, opened 2026-01-14
Allow DS components to implement ServiceFactory/PrototypeServiceFactory instead of service interface

laeubi:
Update SCR specification to support ServiceFactory implementation

- Allow components to implement `ServiceFactory<S>` or `PrototypeServiceFactory<S>` when providing exactly one service
- Create new test bundle tb33 with `ServiceFactory` and `PrototypeServiceFactory` components

Fixes https://github.com/osgi/osgi/issues/688

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ Thumbs.db
 .metadata/
 scrapbook.jpage
 /.recommenders/
+build/
--- a/org.osgi.test.cases.component/bnd.bnd
+++ b/org.osgi.test.cases.component/bnd.bnd
@@ -47,6 +47,7 @@ Import-Package: ${-signaturetest}, *
 	tb30.jar, \
 	tb31.jar, \
 	tb32.jar, \
+	tb33.jar, \
 	tbf1.jar
 	
 -signaturetest = \
--- a/org.osgi.test.cases.component/bnd/tb33.bnd
+++ b/org.osgi.test.cases.component/bnd/tb33.bnd
@@ -1,0 +1,5 @@
+Bundle-Version: 1
+-privatepackage: ${p}.tb33.*
+Service-Component: org/osgi/test/cases/component/tb33/impl/components.xml
+
+
--- a/org.osgi.test.cases.component/src/org/osgi/test/cases/component/junit/DS15TestCase.java
+++ b/org.osgi.test.cases.component/src/org/osgi/test/cases/component/junit/DS15TestCase.java
@@ -820,4 +820,78 @@ public class DS15TestCase {
 					.containsOnly("(osgi.condition.id=" + testName + ")");
 		});
 	}
+
+	/**
+	 * Tests that components can implement ServiceFactory<S> and 
+	 * PrototypeServiceFactory<S> instead of the service interface directly.
+	 * This tests the enhancement from issue #688.
+	 */
+	@Test
+	public void testServiceFactoryImplementation(
+			@InjectInstalledBundle("tb33.jar") Bundle tb33,
+			@InjectService(filter = "(test.name=singleton)", cardinality = 0) ServiceAware<org.osgi.test.cases.component.service.DataService> singletonService,
+			@InjectService(filter = "(test.name=bundle)", cardinality = 0) ServiceAware<org.osgi.test.cases.component.service.DataService> bundleService,
+			@InjectService(filter = "(test.name=prototype)", cardinality = 0) ServiceAware<org.osgi.test.cases.component.service.DataService> prototypeService) throws Exception {
+		
+		assertThat(tb33).as("tb33 bundle").isNotNull();
+		
+		// Start the bundle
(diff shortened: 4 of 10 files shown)


timothyjward:
As was discussed on the call (and in issue #688 where this sort of discussion belongs) it's not clear that this is the right way to solve this. If we do follow this approach then at the very least we would need new interfaces in the SCR package namespace to activate the new behaviour.

timothyjward:
> Just for the sake of "not invented here"?

No, because we're telling SCR to have "special rules" for these interfaces. If SCR needs to enforce special rules then we need an interface which explicitly enables the special rules. Otherwise someone with a component which implements one of these interfaces already will get a breaking change, or if something changes in one of the other interfaces in the future it will break the DS spec.

laeubi:
> I don't see anywhere in the service.component.xml document where you state that the service attribute must be specified when using a factory interface.

Then it has to be made more clear maybe, but it is already in the text because it says 

> If the **component description specifies a service**, the class must

so this must obviously be specified already as `MyService` and we just enhance it that either

> either **implement all interfaces that are provided by the service**, or,
>  if the component provides **exactly one service interface** the class may
> implement `ServiceFactory or `PrototypeServiceFactory`

that is actually quite clear to me.

> We should never expose the ServiceRegistration object to the component.

I don't understand why this should become an issue?

tjwatson:
> > I don't see anywhere in the service.component.xml document where you state that the service attribute must be specified when using a factory interface.
> 
> Then it has to be made more clear maybe, but it is already in the text because it says
> 
> > If the **component description specifies a service**, the class must
> 
> so this must obviously be specified already as `MyService` and we just enhance it that either
> 
> > either **implement all interfaces that are provided by the service**, or,
> > if the component provides **exactly one service interface** the class may
> > implement `ServiceFactory or `PrototypeServiceFactory`
> 
> that is actually quite clear to me.

Not to me, no where do I see it saying that the factory is used as a factory for creating the service object instance ONLY when the `service` attribute is specified by the component.

> 
> > We should never expose the ServiceRegistration object to the component.
> 
> I don't understand why this should become an issue?

Because the `ServiceRegistration` is the domain of the SCR implementation and must never be given to anything outside of the SCR implementation itself.  I've already mentioned why.  There should be no need for the factory to receive the ServiceRegistration anyway if a component providing the factory can only provide a single service type anyway.  The only reason it is needed by the core factory interfaces is so a bundle can use a single factory implementation class to provide service objects to multiple service registrations.  But we will not allow that for SCR service factory components.

It would be much better if the requirements and design were put into an RFC doc so we can review them outside of the spec XML docs and agree to the overall design before diving into the pedantic nature of writing specification text that has to be super precise.

At the call we agreed that a new factory interface is necessary that lives in the domain of the DS specification.  This way we can avoid leaking sensitive objects to the factory, like the ServiceRegistration.

laeubi:
> Not to me, no where do I see it saying that the factory is used as a factory for creating the service object instance ONLY when the `service` attribute is specified by the component.

Can you suggest how we clarify this? As said having a component that *implements* `ServiceFactory` _as a service_ seems purely academic to me, and the context makes it clear. Even if some one would do it it will fall into the first category (implement all services declared). So each component *must* declare (as the name DS suggest) the services and here we deviate from this by allowing it to not implement what is declared but reuse the existing OSGi aproach.

> Because the ServiceRegistration is the domain of the SCR implementation and must never be given to anything outside of the SCR implementation itself. 

It is not required to give *that* [ServiceRegistration](https://docs.osgi.org/javadoc/osgi.core/8.0.0/org/osgi/framework/ServiceRegistration.html)<[S](https://docs.osgi.org/javadoc/osgi.core/8.0.0/org/osgi/framework/ServiceFactory.html)> registration object, as you mention it it not strictly required here and as it only has 3 methods, it would be quite easy for SCR to supply a proxy that redirects some calls (e.g. `getReference()` would be safe) and either ignore or even throw an exception for others.

But as we already have `ComponentContext` we can of course create a `ComponentServiceFactory` or whatever it just looks superfluous to me to came up with that only for the sake of one parameter we can easily protect.

> It would be much better if the requirements and design were put into an RFC doc so we can review them outside of the spec XML docs and agree to the overall design before diving into the pedantic nature of writing specification text that has to be super precise.

I must confess I find it easier to change the spec text so one can discuss concrete changes (as here now). Anyways I already created a ticket that could be seen as an "RFC" and describe what I want https://github.com/osgi/osgi/issues/688 and at least there is one other user that can feel my pain and would find it useful. 

For me this whole bureaucracy might have worked in the past (or not) and often more feels another way to reject a proposal by hoping people will lose interest given all the steps unless something is even discussed, one of the claims a way back when OSGi moved to Eclipse was it will be more open, more dynamic, ... it just don't feels that way, it more feels like better leave the spec away and do whatever you can do on your own or use hacks and workarounds.

PR #922, opened 2026-03-24
Suggested language for the agreed handler chain ordering

timothyjward:
Defines the order for handler chains, including how to position services before or after a statically defined chain.


laeubi (commented):
Just two suggestions, apart from that it looks fine to me.

timothyjward:
I believe that this PR is now complete, including the TCK update. @laeubi - the implementation still requires updates to pass the ordering tests, but I believe that they are now testing the correct handler processing order.

laeubi:
@timothyjward just FYI I found it is currently not easily possible to fulfill the TCK requirement that static + dynamic handlers are merged and get a certain order using the jax-ws RI see:

- https://github.com/eclipse-ee4j/metro-jax-ws/issues/812

I'll try to provide a workaround for the RI case but thought it would be worth noting here as a future reference.

PR #924, opened 2026-03-29
Add requirements document for Java SPI support in the framework

laeubi:
First draft of requirements for core SPI support (issue #372).
This proposes integrating ServiceLoader support directly into the framework's module layer, eliminating the need for
OSGi-specific metadata currently required by the Service Loader Mediator specification.

### DIFF
--- a/.design/requirements-372.md
+++ b/.design/requirements-372.md
@@ -1,0 +1,192 @@
+# Requirements for Java SPI Support in the OSGi Framework
+
+- **Issue**: [osgi/osgi#372](https://github.com/osgi/osgi/issues/372)
+- **Related**: Service Loader Mediator Specification (Chapter 133)
+- **POC Implementations**:
+  [Apache Felix](https://github.com/apache/felix-dev/pull/455),
+  [Eclipse Equinox](https://github.com/eclipse-equinox/equinox/pull/853)
+
+## Terminology
+
+- **Service Loader** -
+  The `java.util.ServiceLoader` API introduced in Java SE 6 for discovering and loading service provider implementations.
+
+- **Service Type** -
+  The interface or abstract class that a Service Provider implements or extends.
+  Identified by its fully qualified class name.
+
+- **Service Provider** -
+  A concrete implementation of a Service Type.
+  Declared in a Provider Configuration File.
+
+- **Provider Configuration File** -
+  A UTF-8 encoded text resource located at `META-INF/services/<ServiceTypeName>` that lists one or more fully qualified names of Service Provider classes, one per line.
+  Lines beginning with `#` are comments.
+
+- **Consumer Bundle** -
+  A bundle that uses the `ServiceLoader` API to discover and load Service Providers.
+
+- **Provider Bundle** -
+  A bundle that contains one or more Provider Configuration Files and the corresponding Service Provider classes.
+
+- **Wired Bundle** -
+  A bundle that is connected to another bundle through the OSGi wiring model, for example via package imports/exports or Require-Bundle.
+
+## Problem Description
+
+Java's `ServiceLoader` API is the standard mechanism for discovering pluggable service implementations on the Java platform.
+Libraries and frameworks across the Java ecosystem rely on `ServiceLoader` for extensibility, including JDBC drivers, JPA providers, XML parsers, logging backends, and many others.
+
+In a standard Java environment, `ServiceLoader` works by scanning the class loader hierarchy for `META-INF/services/` resources and loading the declared implementation classes.
+This assumes a flat or hierarchical class path where all JARs are mutually visible.
+
+In OSGi, each bundle has its own class loader with controlled visibility.
+A bundle's class loader can only find resources within the bundle itself and packages explicitly imported through the wiring model.
+As a result, `ServiceLoader` calls within an OSGi bundle cannot discover Provider Configuration Files or load Service Provider classes from other bundles.
+The `ServiceLoader` API is effectively broken in OSGi without additional intervention.
+
+### Current Solution: Service Loader Mediator
+
+The existing Service Loader Mediator Specification (Chapter 133) addresses this problem through an extender pattern.
+It requires:
+
+1. Provider bundles to declare `osgi.serviceloader` capabilities in their manifest
+2. Consumer bundles to require the `osgi.serviceloader.processor` extender capability
+3. A mediator bundle to be present that processes these capabilities at runtime
+
+While functional, this approach has significant drawbacks:
+
+- **Metadata overhead** -
+  Both provider and consumer bundles require OSGi-specific manifest headers that are not present in standard Java JARs.
+  This forces modification of third-party libraries.
+- **Extender dependency** -
+  The mediator must be deployed and active, adding a runtime dependency and a potential point of failure.
+- **Adoption barrier** -
+  Library authors who are unaware of OSGi or choose not to support it do not include the required capabilities.
+  This is the common case for the vast majority of Java libraries.
+- **Maintenance burden** -
+  Wrapping or patching third-party JARs to add OSGi metadata is fragile and must be repeated for every library version update.
+- **Incomplete integration** -
+  Some advanced `ServiceLoader` usage patterns are difficult or impossible to support through the mediator approach.
+
+### Impact
+
+The inability of `ServiceLoader` to work transparently in OSGi is one of the most frequently cited pain points by OSGi adopters.
+It causes friction when integrating libraries that rely on `ServiceLoader`, which includes a large and growing portion of the Java ecosystem.
+The JDBC, JPA, JAX-RS, JSON-P, JSON-B, and many other Java specifications use `ServiceLoader` for implementation discovery.
+
+## Use Cases
+
+### UC1: JDBC Driver Discovery
+
+A database application bundle imports `java.sql` and calls `DriverManager.getConnection()` or `ServiceLoader.load(Driver.class)`.
+Multiple JDBC driver bundles are installed, each containing a `META-INF/services/java.sql.Driver` file.
+The application bundle should discover all installed JDBC drivers without OSGi-specific metadata in any bundle.
+
+### UC2: Logging Backend Discovery
+
+A library bundle uses SLF4J, which internally calls `ServiceLoader.load(SLF4JServiceProvider.class)`.
+A logging backend bundle (e.g., Logback) is installed with a standard `META-INF/services/` declaration.
+SLF4J should discover the logging backend through normal `ServiceLoader` usage.
+
+### UC3: Standard Java XML Processing
+
+An application calls `DocumentBuilderFactory.newInstance()`, which internally uses `ServiceLoader` to discover XML parser implementations.
+The framework should provide visibility to XML parser providers in other bundles that export the relevant packages.
+
+### UC4: Mixed Environment
+
+Some provider bundles have OSGi-specific `osgi.serviceloader` capabilities (supporting the Mediator spec) while others only have standard `META-INF/services/` files.
+Both styles of provider should be discoverable.
+The core SPI mechanism and the Service Loader Mediator should coexist without conflicts.
+
+### UC5: Module Boundary Enforcement
+
+A consumer bundle imports package `com.example.api` which contains a Service Type.
+Multiple provider bundles also import `com.example.api` and contain Provider Configuration Files.
+A separate bundle that does *not* import `com.example.api` should *not* have its provider configuration files discovered by the consumer.
+The framework must respect wiring boundaries.
+
+### UC6: Dynamic Bundle Lifecycle
+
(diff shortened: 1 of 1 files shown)


.design/requirements-372.md
```diff
@@ -0,0 +1,192 @@
...
+
+An application calls `DocumentBuilderFactory.newInstance()`, which internally uses `ServiceLoader` to discover XML parser implementations.
+The framework should provide visibility to XML parser providers in other bundles that export the relevant packages.
+
+### UC4: Mixed Environment
+
+Some provider bundles have OSGi-specific `osgi.serviceloader` capabilities (supporting the Mediator spec) while others only have standard `META-INF/services/` files.
+Both styles of provider should be discoverable.
+The core SPI mechanism and the Service Loader Mediator should coexist without conflicts.
+
```
HannesWell, line 102: This is an important point in my opinion. Because bundles with Service Loader Mediator  headers will probably be in use for a long time.
But besides the cases where consumer and provider are both using Service Loader Mediator or both are not using it, we also consider and support the case where one of both uses the Mediator.
This could be resolved by assuming that the required headers/wires to 'just' work as expected exists for the participant that's not using the Mediator (actively).

.design/requirements-372.md
```diff
@@ -0,0 +1,192 @@
...
+### UC5: Module Boundary Enforcement
+
+A consumer bundle imports package `com.example.api` which contains a Service Type.
+Multiple provider bundles also import `com.example.api` and contain Provider Configuration Files.
+A separate bundle that does *not* import `com.example.api` should *not* have its provider configuration files discovered by the consumer.
+The framework must respect wiring boundaries.
```
HannesWell, line 108: That's an important point to consider to prevent class-loading/casting issues, also especially with multiple versions of a `...api` bundle available in the runtime.
But that should be feasible since the Framework can simply derive the package name of the Service and can then determine the exporter of that Service interface/class for the currently consuming bundle.

.design/requirements-372.md
```diff
@@ -0,0 +1,192 @@
...
+
+### UC6: Dynamic Bundle Lifecycle
+
+Provider bundles may be installed, started, stopped, and uninstalled at runtime.
+A subsequent `ServiceLoader.load()` call should reflect the current set of resolved and wired provider bundles.
+Previously cached `ServiceLoader` instances are not required to update dynamically, but fresh calls must reflect current state.
```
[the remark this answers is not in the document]
HannesWell: Given that ServiceLoader itself is closed, I assume we cannot do much but consider the bundle state at the time the ServiceLoader reads the `Provider Configuration File` or attempts to load a ServiceProvider class. The latter two actions are the only ones I'm aware of where the ServiceLoader interacts with the Framework and where we can do anything.
E.g. if a Service Provider is loaded from an already uninstalled/stopped bundle, a corresponding Exception could/should be thrown so that the ServiceLoader tries another provider or fails accordingly.

.design/requirements-372.md
```diff
@@ -0,0 +1,192 @@
...
+
+### R1: Transparent SPI Resource Discovery
+
+The framework must enable `ServiceLoader` to discover Provider Configuration Files from provider bundles when the consumer bundle and the provider bundle are both wired to the same Service Type package.
+This must work without requiring any OSGi-specific metadata beyond standard `Import-Package` / `Export-Package` declarations.
+
+### R2: Cross-Bundle Class Loading for SPI Implementations
+
+When `ServiceLoader` attempts to instantiate a Service Provider class declared in a Provider Configuration File from another bundle, the framework must enable loading of that class from the provider bundle that declared it.
+This must work even though the consumer bundle does not explicitly import the package containing the Service Provider class or it is even not exported at all.
```
HannesWell, line 131:
```suggestion
This must work even [if] [...] the package containing the Service Provider class [...] is even not exported at all.
```
If or if not this should be the case, could be a philosophical question.
I think we should also consider if it's possible to load the provider using `Bundle.loadClass()` to keep implementations simpler?

laeubi: > I think we should also consider if it's possible to load the provider using Bundle.loadClass() to keep implementations simpler?

This is effectively the same... just that `Bundle.loadClass()` is a technical detail and I don't want to dictate here how a framework decides it can load the class at all.

.design/requirements-372.md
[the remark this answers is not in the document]
laeubi: > What should happen to the ServiceLoader's iterator?

As pointed out by @HannesWell we do not can (or should) care about the internal structure of the loader and must assume that an SPI is used in a consistent way during a period of time where the system state is stable.

So we can not enforce any dynamics here, if at the time of the call a class is not loadable it will simply throw an error (and Java SPi then does not possibly offer the class through its iterator). Weather or not the iterator is lazy depend on the JVM implementation but one would assume it is.

laeubi:
@timothyjward @HannesWell Anything left here or are your concerns addressed / explained?

.design/requirements-372.md
HannesWell: > The core SPI mechanism and the Service Loader Mediator should coexist without conflicts.

Should this section be enhanced to state that the both mechanisms should also work together? With just coexistence we would have separate world, which IMO wouldn't be good.

HannesWell:
Another question that came to my mind is the case of explicitly set/specified classloaders to be used for ServiceLoader invocations. Until now we just considered the requirements on making the expected service implementations available to the service loader if it's called without explicit classloader and how to make currently invisible implementations visible.
But it might also be the other way round that, for some reason, a specific classloader is passed to a ServiceLoader and that classloader sees only a limited number of jars and services.
In that case it shouldn't happen in an OSGi runtime that the service loader suddenly sees more or different service implementations than in a plain java application.

laeubi:
If a classlaoder different from the bundle classlaoder is passed in then of course it would be used as then the OSGi one is never queried and can not kick in...

.design/requirements-372.md
laeubi: > Should this section be enhanced to state that the both mechanisms should also work together? With just coexistence we would have separate world, which IMO wouldn't be good.

I don't understand what the difference is between "coexist without conflicts" and "work together". At least I do not want to imply any coupling between the both. The core support has to leave alone bundles that use the mediator already (as the should not change) so it just needs to detect the declaration of Mediator Requirements as a bare minimum. This does not mean a framework decides to simply implement *both* at the same time... but this is nothing the specification has to decide.

HannesWell:
> If a classlaoder different from the bundle classlaoder is passed in then of course it would be used as then the OSGi one is never queried and can not kick in...

That's correct for the approach planned for Equinox. But other implementations maybe want to use other techniques like weaving, and then it wouldn't be like that implicitly.
So I think that should be stated.

.design/requirements-372.md
HannesWell: Yes, in general there shouldn't be a coupling between both.

> This does not mean a framework decides to simply implement both at the same time

But for the case when both is implemented I'd recommend some kind of interoperability in order to a smooth transition for 'users' of the mediator spec to the new core requirement.
Because otherwise a provider decides to rely on the new core spec and removes all mediator headers, all consumers would have to be adapted too.

And likewise a consumer that doesn't have the mediator headers will continue to not function if the provider still has the mediator headers.
So I'd add at least the recommendation that, if the Mediator spec is also implemented, both should work together. And yes that would add some coupling.

Alternatively the new core spec could mandate that the mediator spec must not be implemented/adhered at the same, which means effectively all headers from the medaitor spec would have to be ignored.
This would probably be the simpler way to ensure a smooth transition.
I assume that usually the serviceloader headers publish all advertised provided services any ways, so there shouldn't be a difference.

Only specific attribute filters in consumers would be ignored. But I again assume that they are not widely used since it only works in the OSGi world and not in plain Java. And the `ServiceLoader` API suggests other strategies to select a suitable implementation any ways.

laeubi:
I've updated the requirements document to address the feedback:

- **R2**: fixed the wording, and clarified version-safe resolution when multiple versions of the Service Type package are wired (avoids `ClassCastException`s).
- **UC6**: clarified that the framework only acts when a Provider Configuration File is read or a provider class is instantiated; if the provider bundle is no longer wired at that point, an appropriate exception must be thrown so `ServiceLoader` can skip/fail it normally.
- **R11 (new)**: if a caller passes a non-bundle `ClassLoader` to `ServiceLoader.load(...)`, framework SPI support must not expose more than a plain Java application would see — regardless of implementation technique.
- **R12 (new)**: the spec only defines observable behavior, not the technique used to achieve it (stack inspection, weaving, classloader delegation, etc.).
- **R13 (new)**: framework-level SPI support must be configurable via framework properties, at minimum to disable it entirely, and to control its relationship with the Mediator in mixed scenarios (design phase to define specifics).
- **R5**: mixed Mediator/non-Mediator scenarios are intentionally left to the design phase; also clarified the Mediator spec (Ch. 133) is **not** deprecated/replaced, just expected to be superseded over time.
- **R3**: reworded to reference established wires (`BundleWire`/`BundleWiring`) rather than specific headers, so `Require-Bundle` and `DynamicImport-Package` are already covered without extra rules.
- **R1**: made the observable behavior explicit — once SPI support is enabled, any `ClassLoader`-level resource lookup for a Provider Configuration File returns the aggregated view (no `ServiceLoader`-specific detection required); `Bundle#getEntry`/`findEntries` remain available for raw, unaggregated access.

Thanks for the review, @timothyjward and @HannesWell and FYI @stbischof @jjh-aicas @chrisrueger @maho7791

.design/requirements-372.md
```diff
@@ -0,0 +1,208 @@
...
+This specification must be part of the OSGi Core specification as it defines behavior of the framework's module layer class and resource loading mechanisms.
+
+### R11: Explicit Class Loader Behavior
+
+When `ServiceLoader.load(Class, ClassLoader)` (or an equivalent API) is called with a class loader other than a bundle's own class loader, framework-level SPI support must not cause the call to see providers or resources beyond what would be visible in a plain (non-OSGi) Java application using that same class loader.
+This requirement applies regardless of the technique an implementation uses to provide core SPI support (for example, class loader delegation, `ClassLoader` wiring/parenting, or bytecode weaving).
+
+### R12: Implementation Technique Neutrality
+
+This specification must define required, observable behavior only (what a consumer bundle observes when using `ServiceLoader`) and must not mandate the specific technique an implementation uses to achieve that behavior.
```
jjh-aicas, line 199: I do not completely agree to strict technology neutrality.  I require that the solution be usable with bundles that are statically compiled to machine code.  This means that changing the bytecode or using runtime stack tracing are both nogos.  This is why our prototype both limits weaving to modifying the cpool and would allow the weaving to be done before deployment.

laeubi: > I require that the solution be usable with bundles that are statically compiled to machine code
>  This is why our prototype both limits weaving to modifying the cpool and would allow the weaving to be done before deployment

I don't see why this would require something to be declared in the spec, as this would perfectly be valid under the observable behavior requirement to have it woven beforehands.

e.g. you would just transform your bundle and then simply disable any runtime support for SPI (as it can't work under your strict requirements dynamically anyways) - that would be totally valid here.

One could even think about a framework supporting the static weaving during install or deploy time - that's why I think Technique Neutrality is a perfect fit, it does not limit *how* it is archived just what should be observable from the outside.

jjh-aicas: Certainly a means of declaring that weaving was done statically, so that it is not done at runtime is good.  However, with signed bundles, this is not always possible.  The reasons to restrict weaving to cpool changes is that precompilation is still possible.  Static compiled code can get the destination of a call from the cpool, as long as the argument list does not change.

.design/requirements-372.md
```diff
@@ -0,0 +1,208 @@
...
+
+SPI discovery must respect the OSGi wiring model.
+A provider bundle must only be considered for SPI discovery if there is an established wire (as defined by the `org.osgi.framework.wiring` API, e.g. a `BundleWire` between the consumer's and the provider's `BundleWiring`) connecting the consumer bundle to the Service Type package exported by the provider bundle.
+This is independent of the manifest header or resolution mechanism (e.g. `Import-Package`, `Require-Bundle`, or `DynamicImport-Package`) that caused the wire to be established; only the resulting wire, once it exists, is relevant.
+Bundles that have no such established wire for a given Service Type package must not contribute SPI resources or classes.
+
+### R4: No Additional Metadata Requirement
+
+Provider bundles and consumer bundles must not be required to include OSGi-specific manifest headers or capabilities to participate in framework-level SPI discovery.
+Standard Java conventions (`META-INF/services/` files and `Import-Package` declarations) must be sufficient.
```
kun-aicas: Could we extend R4 to cover `module-info.class` as well as` META-INF/services/<ServiceType> ` files?  Supporting only `META-INF/services` would exclude modern modular libraries that use` provides ... with ...` and do not include a duplicate configuration file.

PR #936, opened 2026-05-16
Update RSA TCK to check event exception property

amichair:
In case of an error, the exception instance used to be set in the event property "cause", which was later renamed to "exception", however the tests weren't updated accordingly. This fix checks that both aren't set when there is no error, and "exception" is set when there is one.


timothyjward (approved):

timothyjward:
@stbischof - do we need any further tidy up for this or can it be merged?

stbischof:
Lgtm

PR #937, opened 2026-05-16
Fix leaked services contaminating subsequent tests

amichair:
There are various tests that register services, mainly listeners used in testing generated events. However, some of them are never unregistered, so that they contaminate all subsequent tests. Such leaked listeners pile up to the point of seeing multiple listeners from previous tests still receiving events on later tests. All tests should be isolated, close all of their services and listeners, and run subsequent tests with a clean slate.

### DIFF
--- a/org.osgi.test.cases.remoteserviceadmin/src/org/osgi/test/cases/remoteserviceadmin/junit/RemoteServiceAdminExportTest.java
+++ b/org.osgi.test.cases.remoteserviceadmin/src/org/osgi/test/cases/remoteserviceadmin/junit/RemoteServiceAdminExportTest.java
@@ -98,6 +98,7 @@ public class RemoteServiceAdminExportTest extends DefaultTestBundleControl {
 
 		super.tearDown();
 		ungetAllServices();
+		unregisterAllServices();
 
 		// wait until all remaining exports / imports are gone
 		remoteServiceAdmin = getService(RemoteServiceAdmin.class);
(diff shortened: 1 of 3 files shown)


org.osgi.test.cases.remoteserviceadmin/src/org/osgi/test/cases/remoteserviceadmin/tb2/Activator.java
timothyjward, line 1: This change is incorrect. The event listener is expected to remain registered while tb2 is active (you can see it accessed in `testStop` as part of shutdown https://github.com/amichair/osgi/blob/b2d2d08769ddc15d157906db749d1116adb58a4b/org.osgi.test.cases.remoteserviceadmin/src/org/osgi/test/cases/remoteserviceadmin/tb2/Activator.java#L264). This registration will be automatically removed when tb2 is stopped, and therefore doesn't require explicit cleanup.

org.osgi.test.cases.remoteserviceadmin/src/org/osgi/test/cases/remoteserviceadmin/tb8/Activator.java
timothyjward, line 1: These changes won't break the tests, but as with tb2 this activator is designed to be started and stopped as part of a test. When it is stopped everything that it registers will be cleaned up automatically.

timothyjward (changes requested):

amichair:
ok I removed the activator changes. Unfortunately I don't remember which one of the others is the issue I originally encountered - I saw in test logs that the number of listeners being invoked in each test method was increasing, and printing the instance addresses showed that indeed they were the same instances registered in previous tests. That is to say, there is a real issue here, not just theoretical. I then searched the code for other similar patterns of leaked registrations. I don't know if the solution is correct, but the issue is definitely real and should be addressed.

PR #940, opened 2026-05-17
RSA TCK Listener deduplication and concurrency improvements

amichair:
Improves the event listeners used in the RSA TCK tests:
- Classes were duplicated in each test, fixed by reusing the same common class.
- Concurrency handling seems off: it uses a semaphore for synchronization even though there is no need for managing a number of permits. Further, it seems to get this wrong - adding a permit when adding an event, but then removing anywhere between 0 and 10 permits when an event is removed, so it is unclear what this is trying to achieve. Also, there are multiple waits for timeout/10, possibly an attempt to overcome a bug in concurrency handling here but unclear.

  Another listener just uses plain LinkedList with no synchronization at all.

  This is fixed by reimplementing the common listeners using simple synchronized/wait/notify with more precise control over the concurrency management to make it thread safe with straightforward logic.

### DIFF
--- a/org.osgi.test.cases.remoteserviceadmin/src/org/osgi/test/cases/remoteserviceadmin/common/TestEventHandler.java
+++ b/org.osgi.test.cases.remoteserviceadmin/src/org/osgi/test/cases/remoteserviceadmin/common/TestEventHandler.java
@@ -13,14 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- * SPDX-License-Identifier: Apache-2.0 
+ * SPDX-License-Identifier: Apache-2.0
  *******************************************************************************/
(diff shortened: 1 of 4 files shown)


org.osgi.test.cases.remoteserviceadmin/src/org/osgi/test/cases/remoteserviceadmin/common/TestEventHandler.java
timothyjward, line 1: Synchronising on `this` is poor practice and increases the risk of deadlock in the system as other classes could also synchronise on this object. Using a private internal object (e.g. the list) ensures that we don't end up in that situation.

timothyjward (changes requested):

PR #952, opened 2026-06-08
Incorporate the Jakarta Webservices chapter into the OSGi compendium

timothyjward:
This pull request marks the completion (pending final reviews and any votes) of the Jakarta Web Services Whiteboard chapter.


timothyjward:
This PR is now building properly, including the execution of the tests. Note that:

* The implementation is unable to co-exist with the API bundle as it is only substitutable for the base package. (Not a blocker, just to be aware)
* The implementation fails the handler ordering test `HandlerRegistrationTests#testHandlersOrdering`. This appears to be because the handler DTOs are not in the expected order when the test inspects the runtime dot
* The implementation fails the handler ordering test `HandlerRegistrationTests#testStaticHandlersOrdering`. This appears to be because the handlers aren't ordered properly relative to the static handler chain.

timothyjward:
I believe that we could merge at this point, with the remaining issues to be addressed in the implementation/TCK through bugs.

